PR #11, opened 2019-11-11
Pre-deploy setup

mrchrisadams:
On a call with Jonathan, we had some success setting up, but there's a few things we need to do before we can consider the new admin working

# TODO

- [x] check SQL migrations don't mess with current API
- [x] set up gunicorn to run on correct port
- [x] agree and set up hostname for it, and DNS record
- [x] add the sample .env file, so gunicorn and with these set
- [x] add nice ansible setup to create .env files on server
- [x] add nice ansible setup to create .nginx config files on server
- [x] add nice ansible setup to to nice supervisor workflow for new releases


mrchrisadams:
Hey @jonathan-s  - I think this ansible role will work for you as well, to make it possible to do deploys without relying on me.

Can you message me on zulip to talk about merging this in - it might be better in a short screen sharing session to explain how it's all working.

jonathan-s:
Do you think you could rebase against master, so that commits that are not in master don't show up here?

mrchrisadams:
Ah, nuts, I've messed the history up, when I rebased before.

I can never remember how to fix this, because the PR seems to be showing changes we've already added into master, like the new emails and so on.

jonathan-s:
I believe this should solve the issue. 
```
git checkout ca-add-ansible-for-first-deploy
git rebase master
git push -f origin ca-add-ansible-for-first-deploy
```

mrchrisadams:
OK, I _honestly_ thought I had tried that before, but it seems to have resolved the mess - huzzah!

This should do all the bits listed above PR. 

It relies on you loading env vars yourself, i.e in `pipenv shell` before you run

```
ansible-playbook ansible/deploy.yml -i ansible/inventories/prod
```

To deploy this to a production environment. I've set this up to support staging environments as well.

The dream would be to be able to run automated deploys by pushing to a branch, which triggers ansible runs through git post-receive hooks, but hey, baby steps, no?

ansible/deploy.yml
```diff
@@ -35,18 +38,42 @@
...
+        chdir: "{{ project_root }}/current"
+    - name: Template a file to /etc/files.conf
+      template:
+        src: dotenv.j2
+        dest: "{{ project_root }}/current/.env"
+        owner: deploy
+        group: deploy
+        mode: "0755"
+    - name: collect static files for django
+      shell: "pipenv run ./manage.py collectstatic"
```
jonathan-s: Out of curiosity, where is it set exactly which python version is being run?

ansible/inventories/staging
```diff
@@ -1,2 +1,2 @@
 [staging]
-thegreenwebfoundation.org tgwf_stage="staging" tgwf_domain_name=staging-newadmin project_root="/var/www/{{ tgwf_domain_name }}.thegreenwebfoundation.org"
\ No newline at end of file
+thegreenwebfoundation.org tgwf_stage="staging" tgwf_domain_name=staging-newadmin project_root="/var/www/{{ tgwf_domain_name }}.thegreenwebfoundation.org" gunicorn_port=9003
```
jonathan-s: Do we have separate databases for staging and production, right now?

jonathan-s:
There seems to be a bit cruft from php. But I guess that can be removed at some later point. It would be nice to go through this in screensharing :).

ansible/deploy.yml
mrchrisadams: I believe this is taken care of by the Pipfile, if you list the the version of python in it.

Which we do:

https://github.com/thegreenwebfoundation/greenwebfoundation-admin/blob/master/Pipfile#L33

We have pyenv on the box as well

jonathan-s: You mentioned that it by mistake ran python 2 once, how did that happen?

ansible/inventories/staging
mrchrisadams: Not yet. I'm absolutely in favour of us having one.

A few years back I used this on a project, and I figure it's been in use, so might be worth trying out:

https://github.com/18F/rdbms-subsetter

ansible/deploy.yml
mrchrisadams: I was totally wrong about that. I thought some error I was seeing was because of a change in libraries from py2 to py3, meant that a library was missing that would normally be in py3.

It turned out I hadn't run the pipenv install, and that I should have gone to bed earlier the night before…

PR #13, opened 2019-11-14
Add filter for admin and sending out of email

jonathan-s:
This adds a way to filter for hosts that are shown on the website in admin plus it lets you order by how many ip ranges are registered. (There are several hosts that have 0 ip ranges). 

I've also added the emails that Rene wrote up. 

- [x] Check that emails look ok in console before merging.

### DIFF
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -204,19 +204,37 @@ class HostingAdmin(admin.ModelAdmin):
     def send_email(self, request, *args, **kwargs):
         email_template = request.GET.get('email')
         email_template = f'emails/{email_template}'
+        redirect_name = 'admin:' + get_admin_name(self.model, 'change')
+
         obj = Hostingprovider.objects.get(pk=kwargs['provider'])
-        message = render_to_string(email_template, context={})
+        subject = {
+            'additional-info.txt': 'Additional information needed to approve your listing in the Green Web Directory.',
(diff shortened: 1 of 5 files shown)


mrchrisadams:
@jonathan-s sorry mate, I think me merging in the latest changes for the admin url has caused a conflict here. 

I couldn't tell if this was finished - would you let me know if this needs more work before merging, and when you're ready to merge in request a review so I know to look before merging and redeploying please?

jonathan-s:
I've had a check that the email functionality works. It now does, I couldn't check it prior to this because of a merge conflict. But all good now!

apps/accounts/admin.py
```diff
@@ -161,21 +175,39 @@ def send_button(self, obj):
     send_button.short_description = 'Send email'
 
     def send_email(self, request, *args, **kwargs):
-        email_template = request.GET.get('email')
-        email_template = f'emails/{email_template}'
+        email_name = request.GET.get('email')
+        email_template = f'emails/{email_name}'
+        redirect_name = 'admin:' + get_admin_name(self.model, 'change')
```
mrchrisadams: I've seen three different ways of formatting strings in the last three pull requests, and it's not stopping functionality, but it is a bit inconsistent. I've created an issue to agree on style for this reason, to discuss in future,

https://github.com/thegreenwebfoundation/greenwebfoundation-admin/issues/23

jonathan-s: Good point.

PR #19, opened 2019-11-19
Amend hostingproviders_stats so that we can remove hostingproviders in admin

jonathan-s:
Note to reader: deleting a hostingprovider checks all relations, which means deleting one provider is fairly slow due to the huge greencheck table.

### DIFF
--- a/apps/accounts/management/commands/amend_hostingproviders_stats.py
+++ b/apps/accounts/management/commands/amend_hostingproviders_stats.py
@@ -1,0 +1,31 @@
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+
+class Command(BaseCommand):
+    help = "Add missing id column for hostingstats."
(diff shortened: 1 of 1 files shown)


mrchrisadams:
Hi Jonathan

If I understand this correctly, this migration:

- makes a new a new table called `hostingproviders_stats_copy`
- copies the content across from `hostingproviders_stats`
- deletes the old table
- renames the new one

jonathan-s:
There are no caveats to this that I can think of, and it does exactly what you just described. Perhaps that this should be run in a transaction. I can make this amendment to the code.

mrchrisadams:
Could you mind making that change please for adding the transaction? Once I have that, I'll carry out the deploy

jonathan-s:
@mrchrisadams I've made that change.

PR #24, opened 2019-11-21
Add timestamps to ranges and approvals + migration

jonathan-s:
The current database does not contain when an ip range, as range or an approval was created. It doesn't exist in those tables. So we add a migration that makes it possible to view when an entry is created or modified. 

Whenever you create something in symfony it is logged in the table, `ext_log_entries`, there we can extract the date. 

So the management command picks the latest date that was stored and applies that as the creation date. 

This will help rene to see when an approval or a range was created, and how old it is.


mrchrisadams:
Sorry mate, I'm not sure what's going on here.

Can you add a bit of explanation to the pr so I know what to look for, and what the intended behaviour is?

jonathan-s:
@mrchrisadams I updated the description, let me know if anything is unclear.

mrchrisadams:
Ah, this is _much_ better!

Thanks. Is there any order these PRs need to be merged in an run BTW? I'm likely going to batch them up to deploy tomorrow morning.

jonathan-s:
They are completely independent, so any order will do.

PR #31, opened 2019-11-25
Be able to check an url in admin

jonathan-s:
Adds a view in admin so that you can check urls there as well. It was a request from Rene.

### DIFF
--- a/apps/accounts/admin_site.py
+++ b/apps/accounts/admin_site.py
@@ -1,5 +1,54 @@
+import re
+
 from django.contrib.admin.sites import AdminSite
 from django.contrib.auth.forms import AuthenticationForm
+from django.core.exceptions import ValidationError
+from django.urls import path
+from django.urls import reverse
+from django.shortcuts import render
+
+import requests
+from requests.exceptions import HTTPError
+
+from django.views.generic.edit import FormView
+from django import forms
+
+
+URL_RE = re.compile(r'https?:\/\/(.*)')
+BASE_URL = 'https://api.thegreenwebfoundation.org/greencheck'
+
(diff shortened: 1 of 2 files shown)


apps/accounts/admin_site.py
```diff
@@ -1,5 +1,48 @@
...
+class CheckUrlForm(forms.Form):
+    url = forms.URLField()
+
+    def check_green(self):
+        url = self.cleaned_data['url']
+        cleaned_url = URL_RE.match(url).group(1)
+        resp = requests.get('{base_url}/{url}'.format(
+            base_url=BASE_URL, url=cleaned_url)
+        )
+        return resp.json().get('green', False)
```
mrchrisadams: If the response is a bad one, will calling json() here cause an Exception, or just return `None` or a similar object?

If there's a chance it'll return a bad result, that `json()` or `get()` chokes on, would you mind adding some code to catch the exception by [throwing a big tarp][1] or similar?

[1]: https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/

mrchrisadams: Also, can we settle on using f-strings as our default way to interpolate variables in strings?

We've used them in a few places, and I think they'd remove some redundant assignment of values to variables that we have here with the `format()` approach.

jonathan-s: Ok, we'll settle for f strings. I added a validation step, so if the server for some reason bails out it will show a nice error too.

mrchrisadams: This looks good, thanks @jonathan-s

PR #33, opened 2019-12-03
Create procedures with management command

jonathan-s:
This PR does the following
- Add the green_presenting table. URLs are unique. 
- Create a procedure for inserting urls into green_presenting. If the inserted url is not green we try to delete it. If the url is green we try to insert it. If it already exists (urls are unique, we update the date instead). 
- A backfill procedure to run to backfill the table `green_presenting`. 
- A trigger for the greencheck table, once the trigger is active it will push new urls into green_presenting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,6 +125,7 @@ dmypy.json
 notes.md
 sublime.sublime-workspace
 static/*.*
+static
 
 # do not track ansible cruft
 deploy.retry
--- a/apps/greencheck/management/commands/create_greencheck_trigger.py
+++ b/apps/greencheck/management/commands/create_greencheck_trigger.py
@@ -1,0 +1,17 @@
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+
+class Command(BaseCommand):
+    help = "Create trigger to insert urls."
+
+    def handle(self, *args, **options):
+        with connection.cursor() as cursor:
+            cursor.execute('''
+                -- CREATE TRIGGER "after_greencheck_insert" ---------------
+                CREATE TRIGGER after_greencheck_insert
+                    AFTER INSERT ON greencheck
+                    FOR EACH ROW
+                CALL insert_urls(NEW.url, NEW.green, NEW.id_hp, NEW.datum);
+                -- --------------------------------------------------------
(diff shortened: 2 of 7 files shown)


mrchrisadams:
Hi @jonathan-s , this looks good to me.

If I understand this, we have two management commands here:

- **insert_procedures** - which does the following:
  - creates the table
  - defines a procedure for updating the presenting table
  - runs the backfill procedure for the huge greencheck table
- **create_greencheck_trigger** which sets up the trigger to update the `green_presenting` table

I'm happy to run this, but given we'll only run these commands once, was there a specific reason you chose  to have a management command over having a data migration instead to do the same work?

I don't have strong feelings each way, but I'm mindful that one of these management commands will likely end up needing to run for 3-4 days, and I don't think it's the kind of thing we want to accidentally trigger in future afterwards. 

To be clear, I'm happy with this in its current shape, but I'll be on a train most of tomorrow so if you merge it and run the commands, I won't be able to help troubleshoot anything until the late afternoon at the earliest, if there some was something we didn't consider.

Can you let me know when you're around tomorrow for a short call? If you're free between 4:30 and 6, it might be worth doing a short screening session to kick this off…

jonathan-s:
I'll be free sometime around 4:30 and 6pm. So I'd be happy to have a call then. So both of these management commands is only for creation of the procedures, and you're right, perhaps it would have been better to have it in a migration. At least for `insert_procedures`. 

With regards to `create_greencheck_trigger`. As soon as that command is run we **expect** that `insert_procedures` has already run. From then on every insertion in greencheck will create an entry in `green_presenting`. 

In short all of these management commands will run in a snap. You'll have to run a sql statement `CALL backfill();` to actually create the backfilling of the entire table. That sql statement will take around 3 days to finish.

mrchrisadams:
Ah, thanks for clearing up those last bits about it _defining_ a procedure, not _calling_ the procedure :)

I have no concerns about running these now - but it would be good for a quick check in anyway, as your approach here means that creating snapshots of that domain table is much simpler than the earlier plan we had, and it's worth a quick discussion, about the next steps about creating a nice data pipeline for publishing these datasets 👍

PR #37, opened 2019-12-09
Some improvements for the procedure

### LINKED ISSUE #36: Parameterise the domain backfill procedure for our green domains list
After a call with @jonathan-s, we agreed that being able to pass in a date for the current backfill procedure would make it easier to test, and use in production.

So rather than simply having us call:

```sql
CALL backfill()
```

We'd be able to do:

```sql
CALL backfill("2019-12-01")
```

And have this only fill domains added to the greencheck table since the date.

This lets us make snapshots more quickly, and would bring down the expected time for a backfill run from 3.5 days for the last ten years, to a few minutes if we only care about the last month or so.

jonathan-s:
We discussed in #36 what needed to be done. So after this is merged it will close #36 

To get this updated, all that you should need to do would be to run the management command `insert_procedures` again

### DIFF
--- a/apps/greencheck/templates/backfill_procedure.sql
+++ b/apps/greencheck/templates/backfill_procedure.sql
@@ -8,7 +8,7 @@ BEGIN
   DECLARE ggreen TINYINT(2);
   DECLARE gid_hp INT(11);
   DECLARE gurl VARCHAR(255);
-  DECLARE cur CURSOR FOR SELECT datum, green, id_hp, url FROM greencheck where id_hp > 0 AND datum > date_begin;
+  DECLARE cur CURSOR FOR SELECT datum, green, id_hp, url FROM greencheck WHERE id_hp > 0 AND datum > date_begin;
   DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
 
   OPEN cur;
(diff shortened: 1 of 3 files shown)


apps/greencheck/templates/backfill_procedure.sql
```diff
@@ -1,14 +1,14 @@
...
+CREATE PROCEDURE backfill(IN date_begin Date)
 BEGIN
   DECLARE done INT DEFAULT FALSE;
 
   DECLARE gdatum DATETIME;
   DECLARE ggreen TINYINT(2);
   DECLARE gid_hp INT(11);
   DECLARE gurl VARCHAR(255);
-  DECLARE cur CURSOR FOR SELECT datum, green, id_hp, url FROM greencheck where id_hp > 0;
+  DECLARE cur CURSOR FOR SELECT datum, green, id_hp, url FROM greencheck where id_hp > 0 AND datum > date_begin;
```
mrchrisadams: it's a small thing, but can we please have all SQL statements in caps, to more easily differentiate from variables? i.e. WHERE being in caps too?

More here - https://www.sqlstyle.guide/#reserved-words

If possible, for future raw SQL that we write, I'd be happier, where we won't have existing conventions on the project to follow the SQL style outlined in this guide, for readability:

https://www.sqlstyle.guide

mrchrisadams:
To be clear, once this style issue is addressed, and you have confirmed the other behaviour, please merge in and deploy.

apps/greencheck/templates/backfill_procedure.sql
jonathan-s: Good call.

jonathan-s:
I've confirmed that I'm seeing entries in `green_presenting` during the execution of `backfill`.

PR #38, opened 2019-12-09
Dump the data + Present where you can download it

### LINKED ISSUE #35: Dump table that contains all green urls & upload to object storage
As #33 is now merged and we've got the procedures in place we are now able to create the SQLite database from the table `green_presenting`. We can use https://github.com/simonw/db-to-sqlite that creates a sqlite file from a table without too much fuss. 

Once we've got the sqlite file we make sure that this file is uploaded to the object storage. 

The last step is to be able to present all the uploaded objects to the user. We can do that by creating a page in admin that is accessible without login. This would page would query the object storage and display the 10 most recent uploaded sqlite files.

jonathan-s:
So this PR does two things
* Creates a management command that dumps the data from the presenting table. 
* Creates a view where you can view all recent datadumps. 

For production we need to create to add the environment variable `GOOGLE_APPLICATION_CREDENTIALS` together with the path to the json credential. 

This PR closes #35

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -28,6 +28,9 @@ gunicorn = "==19.9.0"
 mysqlclient = "==1.4.4"
 requests = "==2.22.0"
 whitenoise = {extras = ["brotli"],version = "==4.1.4"}
+db-to-sqlite = "==1.0.2"
+sqlite-utils = "==1.12.1"
+google-cloud-storage = "==1.23.0"
 
 [requires]
 python_version = "3.6"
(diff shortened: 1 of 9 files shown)


mrchrisadams:
hi @jonathan-s , thanks for this. 

Can you please change the management command, to accept a flag, so it's possible to export the sqlite file _without_ immediately uploading to google cloud storage?

My use case is taking a snapshot that can then be inspected, or served using something like datasette on a single server, without needing to fetch it from object storage first.

Thanks

jonathan-s:
I've added a flag now :)

PR #44, opened 2020-01-10
Filter non-staff users so we don't leak user data

### LINKED ISSUE #41: A user that only is active should not be able to see other users
They should only be able to see their own user, not other users. We don't want to leak that data.

jonathan-s:
Closes #41


mrchrisadams:
Thanks @jonathan-s  👍

PR #46, opened 2020-01-14
WIP - Update data cleaning scripts

mrchrisadams:
This PR adds a slightly different approach for backfilling the data into the green urls data, in the `green_presenting` table.

This consists of

- a new stored procedure, `add_latest_check_for_url` to add the latest result for a given url from the `greencheck` table. This looks for the latest check in backwards, rather than going through all urls, over the last N years until we find the latest check. Save work with a transaction when we _do_ find a url, so we can resume if need be. 
- second stored procedure, `backfill_by_url` to loop through a set of urls in a table, using this new `add_latest_check_for_url` procedure
- some python functions for cleaning a list distinct urls we might get from the main greencheck table, to strip out invalid hostnames - I used this when checking different sets of urls. About 23k out 20million urls were badly formed, and they caused the backfill procedures to choke.

### DIFF
--- a/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
+++ b/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
@@ -22,6 +22,17 @@ BEGIN
   DECLARE h_hostpartner     VARCHAR(255);
   DECLARE h_hostwebsite     VARCHAR(255);
 
+
+  -- SELECT concat('** checking for: ', purl) AS '** DEBUG:';
+
+  DECLARE CONTINUE HANDLER FOR NOT FOUND
+    BEGIN
+      -- SELECT concat('** NO RESULT FOR: ', purl) AS '** DEBUG:';
+      -- SELECT concat('** BUT WE SHOULD NOT BUBBLE THE EXCEPTION UP: ', '') AS '** DEBUG:';
+    END;
+
+
+
   -- fetch the latest result for this url, using the index
   SELECT id, id_hp, id_greencheck, type, url, ip, datum, green, tld
   FROM greencheck
@@ -38,6 +49,8 @@ BEGIN
 
     -- check
     IF g_green = 'yes' THEN
+
+      -- SELECT concat('** GREEN: ', purl) AS '** DEBUG:';
       INSERT INTO green_presenting
       (
         `id`,
--- a/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
+++ b/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
@@ -50,6 +50,7 @@ BEGIN
     -- check
     IF g_green = 'yes' THEN
 
+      START TRANSACTION;
       -- SELECT concat('** GREEN: ', purl) AS '** DEBUG:';
       INSERT INTO green_presenting
       (
@@ -73,5 +74,8 @@ BEGIN
         `hosted_by_id` = g_id_hp,
         `hosted_by_website` = h_hostwebsite,
         `partner` = h_hostpartner;
+
+      COMMIT;
+
     END IF;
 END
--- a/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
+++ b/apps/greencheck/templates/add_latest_check_for_url_procedure.sql
@@ -1,0 +1,64 @@
+DROP PROCEDURE IF EXISTS add_latest_check_for_url;
+
+-- define our procedure, that accepts the url param, as a VarChar
+CREATE PROCEDURE add_latest_check_for_url(IN purl VarChar(255))
+
+BEGIN
+
+
+  -- declare the types we'd fetch from the cursor
+  DECLARE g_id            BIGINT(20);
+  DECLARE g_id_hp         INT(11);
+  DECLARE g_id_greencheck INT(11);
+  DECLARE g_type          ENUM('url','whois','ip','none','as');
+  DECLARE g_url           VARCHAR(255);
+  DECLARE g_ip            DECIMAL(39,0);
+  DECLARE g_datum         TIMESTAMP;
+  DECLARE g_green         ENUM('yes','no','old');
+  DECLARE g_tld           VARCHAR(64);
+
+  -- declare the other types for the next query
+  DECLARE h_hostname        VARCHAR(255);
+  DECLARE h_hostpartner     VARCHAR(255);
+  DECLARE h_hostwebsite     VARCHAR(255);
+
+  -- fetch the latest result for this url, using the index
+  SELECT id, id_hp, id_greencheck, type, url, ip, datum, green, tld
+  FROM greencheck
+  WHERE url = purl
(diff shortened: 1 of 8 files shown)


apps/greencheck/templates/add_latest_check_for_url_procedure.sql
```diff
@@ -0,0 +1,81 @@
...
+  DECLARE h_hostwebsite     VARCHAR(255);
+
+
+  -- SELECT concat('** checking for: ', purl) AS '** DEBUG:';
+
+  DECLARE CONTINUE HANDLER FOR NOT FOUND
+    BEGIN
+      -- SELECT concat('** NO RESULT FOR: ', purl) AS '** DEBUG:';
+      -- SELECT concat('** BUT WE SHOULD NOT BUBBLE THE EXCEPTION UP: ', '') AS '** DEBUG:';
+    END;
```
jonathan-s: What does the following five lines of code do?

jonathan-s: In what way is this connected to line 23 in backfill_by_url_procedure?

apps/greencheck/templates/backfill_by_url_procedure.sql
```diff
@@ -0,0 +1,51 @@
...
+  -- we need a way to check so we can exit the loop
+  DECLARE done INT(8) DEFAULT FALSE;
+  DECLARE counter INT(8) DEFAULT FALSE;
+
+  -- declare the types we'd fetch from the cursor
+  DECLARE fetched_url                VARCHAR(255);
+
+  -- set up our cursor, so we have a list of urls to loop through
+  DECLARE cur CURSOR FOR
+  SELECT url FROM urls;
```
jonathan-s: From what I understand the urls table is a temporary table that you've filled from a csv. From where does that CSV come from? Is it a dump of all urls?

apps/greencheck/url2green.py
```diff
@@ -0,0 +1,45 @@
...
+# and only write the lines that count as real hostnames
+# we can check the host
+
+
+def lazy_messy_url_list(path_to_infile):
+    """
+    Accept a text file at `path`, with one url per line,
+    and return an generator to iterate through each line
+    """
+    with open(path, "r") as messy_list_of_urls:
```
jonathan-s: should this not be `path_to_infile`?

apps/greencheck/url2green.py
```diff
@@ -0,0 +1,45 @@
...
+    if len(hostname) > 255:
+        return False
+    if hostname[-1] == ".":
+        hostname = hostname[:-1] # strip exactly one dot from the right, if present
+    allowed = re.compile("(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
+    return all(allowed.match(x) for x in hostname.split("."))
+
+# Usage
+# urls = lazy_messy_url_list("path/to/file")
+# write_clean_urls(urls)
```
jonathan-s: Possibly just write a function that does the above? That would show the use case as well if you'd need to decompose it for whatever reason.

jonathan-s:
Neat! Would you say that this makes the other scripts obsolete? This has the advantage of not checking all urls (ie speed). There is a chance that it will classify a url as green if it isn't. (This is an edge case when you have a url that first was green and then became grey and stayed that way). I'm happy with this tradeoff, but it's worth keeping in mind :).

apps/greencheck/templates/backfill_by_url_procedure.sql
mrchrisadams: It's a dump of all the urls, once I've filtered out the ones that weren't valid hostnames.

I did indeed create a temporary table for this.

My thinking is that it lets us either parallelize the work, or be able to pick up where we left off.

apps/greencheck/templates/add_latest_check_for_url_procedure.sql
mrchrisadams: As far as I can tell, stored procedures bubble `NOT FOUND` results up to the calling stored procedure.

So the `backfill_by_url` procedure was dropping out of the loop early, as soon as a for a url didn't return any results.

This meant that green presenting wasn't populated at all if I didn't catch it.

mrchrisadams:
Gah. Even with these changes, running these stored procedures still takes aaaages.

We have 20 million urls to update info for.

Running locally, I get this timing - I've been leaving this to run in the background today.

I was worried that we'd see the same awful numbers for the first 10000 as for the rest, but it at least speeds up after the first few hundred thousand. Given that this is checking against a greencheck table of around 80 m rows locally, and on prod we have 1.5bn I still think we might need another approach to get this done in a sensible time, as on urls that get lots of checks, like yahoo.com, we the stored procedure still takes *minutes* .




```
mysql me@localhost:greencheck> CALL backfill_by_url;

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 100000 |
+--------------------+
1 row in set
Time: 2652.390s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 200000 |
+--------------------+
1 row in set
Time: 4876.233s
+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 300000 |
+--------------------+
1 row in set
Time: 1600.736s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 400000 |
+--------------------+
1 row in set
Time: 2556.080s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 500000 |
+--------------------+
1 row in set
Time: 119.215s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 600000 |
+--------------------+
1 row in set
Time: 175.564s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 700000 |
+--------------------+
1 row in set
Time: 87.589s

+--------------------+
| ** DEBUG:          |
+--------------------+
| ** progress 800000 |
+--------------------+
1 row in set
Time: 214.263s

```

apps/greencheck/templates/backfill_by_url_procedure.sql
jonathan-s: If it is a dump of all the urls, are they unique? If they are we'll reduce the amount of urls. Perhaps we could reduce the space even further by only including one url with the domain. For instance

- sub.example.com
- sub2.example.com
- www.example.com
- www.example.com/hello

This would only show up once in that table.

jonathan-s: Alternatively we could populate the url table with this -> https://gist.github.com/chilts/7229605 and then just rely on the stored procedure that automatically inputs new urls in the green_presenting table. 

After all "Perfect is the enemy of good", and that will get us a green2url zip that we can slowly expand upon over time.

apps/greencheck/url2green.py
mrchrisadams: Good call. 👍

PR #49, opened 2020-01-24
Add hostingprovider and datacenter group to new users

jonathan-s:
We add the hostingprovider and the datacenter group to new users. I also saw that we had hardcoded the success url so that is now removed.


mrchrisadams:
Thanks for this @jonathan-s.

PR #53, opened 2020-02-01
Inherit from field instead of DecimalField

### LINKED ISSUE #51: 500 triggered when I try try to update my hosting provider
On production, when I try to update hosting provider details for an existing account, I get a 500.

Screenshot 2020-01-29 at 19 45 46

I've shared the stack trace, but it should also be visible in sentry:

https://sentry.io/share/issue/f9d309a6e4804c258805ae3d0cfab221/

I think I can reproduce this by creating a new hosting provider for a website I own.

Try signing in and creating an IP entry for your own website. I think is triggered when accessing the change view for a given hosting provider.

## Steps to reproduce

- sign in as user
- hit create for hosting provider
- fill in details
- hit submit
- 😞

jonathan-s:
The underlying issue was that when admin rendered the field. We had the following function being executed. 

```
def display_for_field(value, field, empty_value_display):
    from django.contrib.admin.templatetags.admin_list import _boolean_icon

    if getattr(field, 'flatchoices', None):
        return dict(field.flatchoices).get(value, empty_value_display)
    # BooleanField needs special-case null-handling, so it comes before the
    # general null test.
    elif isinstance(field, models.BooleanField):
        return _boolean_icon(value)
    elif value is None:
        return empty_value_display
    elif isinstance(field, models.DateTimeField):
        return formats.localize(timezone.template_localtime(value))
    elif isinstance(field, (models.DateField, models.TimeField)):
        return formats.localize(value)
    elif isinstance(field, models.DecimalField):
        return formats.number_format(value, field.decimal_places)
    elif isinstance(field, (models.IntegerField, models.FloatField)):
        return formats.number_format(value)
    elif isinstance(field, models.FileField) and value:
        return format_html('<a href="{}">{}</a>', value.url, value)
    else:
        return display_for_value(value, empty_value_display)
```

Since IpAddress inherited from DecimalField we stumbled upon this issue. It would be great if you can test this before merging this in. 

What I've tested. In a created hostingprovider, I've successfully created a new IP without any error. 

Fixes #51

### DIFF
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -220,11 +220,10 @@ class HostingAdmin(admin.ModelAdmin):
             'Email sent to user'
         )
 
-        # This table is unusable right now.
-        # HostingCommunication.objects.create(
-        #     template=email_template,
-        #     hostingprovider=obj
-        # )
+        HostingCommunication.objects.create(
+            template=email_template,
+            hostingprovider=obj
+        )
 
         name = 'admin:' + get_admin_name(self.model, 'change')
         return redirect(name, obj.pk)
--- a/apps/greencheck/models.py
+++ b/apps/greencheck/models.py
@@ -1,12 +1,16 @@
 import decimal
 import ipaddress
 
-from django.utils.text import capfirst
 from django import forms
 from django.db import models
+from django.db.models.fields import Field
 from django_unixdatetimefield import UnixDateTimeField
 from django_mysql.models import EnumField
 from django.core import exceptions
+from django.core import validators
+from django.utils.text import capfirst
+from django.utils.functional import cached_property
+
 from model_utils.models import TimeStampedModel
 
 from apps.accounts.models import Hostingprovider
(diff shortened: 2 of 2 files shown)


mrchrisadams:
Hi Jonathan,

I don't have a user set up on a local database with the correct perms for this, but I'll have a chance at lunchtime to test this.

I'd _really_ like some tests for stuff like this in future, as it's hard to me to see why this fixes the problem without the same context you now have.

If you can't add tests, can you at least link to some documentation that explains what's happening, and what the value was that was triggering the error here in your case?

jonathan-s:
@mrchrisadams I can't really link to some documentation, as I had to dig into the code to figure out what was happening. I'll try to explain the case through linking to some code. 

So for an admin read only field it starts to try to display the contents of the field here: https://github.com/django/django/blob/master/django/contrib/admin/helpers.py#L218

Peeking into the function `display_for_field`. We see that it determines how to display that field by checking what class the field was from -> https://github.com/django/django/blob/1a09708dcb2f60ad5ca0a75b8de9619356f74ff6/django/contrib/admin/utils.py#L395

Since before this patch we inherited from a `DecimalField` it now tries to convert the value (in this case `127.0.0.1`) into a Decimal. Since it can't do that it abruptly fails. 

As you can see from `display_for_field` it has a fallback that just does `str(value)` if it can't find a matching case (ie when we are inheriting from `Field`). This is what we want.

PR #57, opened 2020-02-06
Fix showonwebsite error

mrchrisadams:
When people who are not staff members try to add datacentres in the new admin, was throwing an error.

This is because the show on website value was being passed as 'None'.


jonathan-s:
I think the better approach would be to change this -> https://github.com/thegreenwebfoundation/greenwebfoundation-admin/blob/master/apps/accounts/models/hosting.py#L31 and add default=False

mrchrisadams:
Thanks @jonathan-s . Have updated.

jonathan-s:
Great! I'll need the migration as well. So `python manage.py makemigrations` should do it

mrchrisadams:
I left out the migration initially because it doesn't result in any material changes. Here's the SQL output

```
BEGIN;
--
-- Alter field showonwebsite on datacenter
--
COMMIT;
```

Might as well leave it in til we squash migrations later, I guess.

jonathan-s:
I thought it was going to have database changes as well, oh well. Thanks for leaving in the migration, it's always a bit confusing if you've got hanging migrations.

PR #59, opened 2020-02-06
Add missing mixin to allow ASN submissions

mrchrisadams:
Right now, non-staff users get a 500 when they try to update info about an ASN.

This make it possible for them to make requests with ASN numbers again


mrchrisadams:
Merging this in as:

- it's so small
- I see continuing errors in sentry
- the current behaviour triggers a 500

jonathan-s:
All good!

PR #60, opened 2020-02-06
Fix hosting provider deletion

mrchrisadams:
This PR fixes the issue reported in Sentry when a user tries to delete a hosting provider.

Because:

1. there is foreign key from _Hosting Provider Stats_ to _Hosting Provider_.
2. the primary key HostigProviderStats is actually `id_hp`,

We end up in a situation where django, when deleting a hosting provider will try to follow the foreign key relation to delete the corresponding set of hosting provider stats. By default it looks for a column named `id`, which it assumes to be the primary key for the column. In our case it's `id_hp`.

P.S. Before today, I had never seen Foreign Keys that also act as Primary Keys. Mind blown.

```
mysql chris@localhost:greencheck> DESCRIBE hostingproviders_stats
+---------------+---------+------+-----+---------+-------+
| Field         | Type    | Null | Key | Default | Extra |
+---------------+---------+------+-----+---------+-------+
| id_hp         | int(11) | NO   | PRI | <null>  |       |
| green_checks  | int(11) | NO   |     | <null>  |       |
| green_domains | int(11) | NO   |     | <null>  |       |
+---------------+---------+------+-----+---------+-------+
3 rows in set
Time: 0.029s

mysql chris@localhost:greencheck> SELECT * FROM hostingproviders_stats LIMIT 2
+-------+--------------+---------------+
| id_hp | green_checks | green_domains |
+-------+--------------+---------------+
| 1     | 2406         | 40            |
| 3     | 2268112      | 40957         |
+-------+--------------+---------------+
2 rows in set
Time: 0.039s

```


jonathan-s:
Nuts! Migration is missing though.

mrchrisadams:
OK, I've added the migration.

It's really important to fake this one. the SQL generated will try to drop a column that isn't there, so when we run it  on production we'll need to `--fake` it.

jonathan-s:
Ok in that case it might be better to store these changes in a prior migration that has already run. Ideally the very first one.

apps/accounts/migrations/0008_auto_20200206_1737.py
```diff
@@ -0,0 +1,27 @@
...
+            options={'verbose_name': 'Hosting Provider'},
+        ),
+        migrations.RemoveField(
+            model_name='hostingproviderstats',
+            name='id',
+        ),
+        migrations.AlterField(
+            model_name='hostingproviderstats',
+            name='hostingprovider',
+            field=models.ForeignKey(db_column='id_hp', on_delete=django.db.models.deletion.CASCADE, primary_key=True, serialize=False, to='accounts.Hostingprovider'),
```
jonathan-s: So it might be better to change this to primary_key=True when this model was created.

PR #61, opened 2020-02-07
Approved should be set by default

jonathan-s:
An error was recently triggered -> https://sentry.io/organizations/product-science/issues/1501302817/?project=2071451&referrer=alert_email

Approved should be set by default here. I checked that the migration did not create any database changes by doing sqlmigrate. 

So there we can make these changes in the initial migration to avoid migration bloat.


jonathan-s:
As a side note. I don't we expose the intermediary table in admin. Perhaps we should be doing that too.

mrchrisadams:
Thanks. Merging this and deploying

PR #63, opened 2020-03-09
Correct signature for check url

jonathan-s:
The definition signature needs to accept kwargs as seen in the issue on sentry -> https://sentry.io/share/issue/ec0022e3308d4f6cad10856d88794db1/

See also: https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/views/generic/edit.py#L63

### DIFF
--- a/apps/accounts/admin_site.py
+++ b/apps/accounts/admin_site.py
@@ -41,8 +41,8 @@ class CheckUrlView(FormView):
     form_class = CheckUrlForm
     success_url = '/not/used'
 
-    def get_context_data(self):
-        context = super().get_context_data()
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
         context['form_url'] = reverse('admin:check_url')
         return context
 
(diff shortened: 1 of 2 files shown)


mrchrisadams:
@jonathan-s  - can you give me some more context on what this fixes, and might have been broken before I merge this in mate?

I can see some info from that error, but I don't have any issue or error to look to see if this fixes anything, so I can't test the expected behaviour.

I'm currently able to try urls here, so it doesn't seem to be that.

https://admin.thegreenwebfoundation.org/try_out/

jonathan-s:
@mrchrisadams If you try out this url it will fail `https://oeuou.com$/rom€#&"` it passes the html url check but doesn't pass the backend url validity check. 

Line 61 https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/views/generic/edit.py#L61 

requires a form argument which it would take from kwargs. Since before this we didn't have kwargs it fails.

mrchrisadams:
Ah, thanks. Merging in 👍

PR #64, opened 2020-03-21
Add importer for publicly listed IP addresses for AWS

mrchrisadams:
This PR adds a few things

### Add tests with pytest

We now have a test harness for future development. I needed this to catch a hard to diagnose bug with the new IpAddress field implemented earlier.

### Add an management command for importing AWS public IP ranges

We also have management command to run imports from public listings of IP ranges, as listed in issue 18 on the green web foundation.

https://github.com/thegreenwebfoundation/thegreenwebfoundation/issues/18

This updates the 5 regions listed as green/sustainable by Amazon, and imports their IPs.

### Things to bear in mind before we deploy this and run the import

*Overlapping IP ranges now that we have data direct from AWS* - We've seen bugs already where the API has been expecting just a single response back, when checking if a domain has an IP inside a given IP address. You can see this documented in issue 48 on the TGWF repo as described below:

https://github.com/thegreenwebfoundation/thegreenwebfoundation/issues/48

If this is the case, we'll need to account for this before we merge it, otherwise we might trigger a bunch of 500's.

@arendjantetteroo - I know that before running this import, some people who are already customers of Amazon, have listed their IP ranges with us. This means we'd end up with two hosters being listed. How do we handle cases when we have multiple hosters returned?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: python
 python:
-  - "3.6"      # current default Python on Travis CI
+  - "3.6"
 
-# command to install dependencies
 install:
   - pip install pipenv --upgrade-strategy=only-if-needed
   - pipenv install --dev
 # command to run tests
 script:
   - pytest
+services:
+    - mysql
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+  - "3.6"      # current default Python on Travis CI
+
+# command to install dependencies
+install:
+  - pip install pipenv --upgrade-strategy=only-if-needed
+  - pipenv install --dev
+# command to run tests
+script:
+  - pytest
--- a/Pipfile
+++ b/Pipfile
@@ -13,6 +13,9 @@ ipython = "*"
 ipdb = "*"
 mycli = "*"
 werkzeug = "*"
+pytest = "*"
+pytest-django = "*"
+pytest-watch = "*"
 
 [packages]
 # psycopg2-binary = "*", not needed for now.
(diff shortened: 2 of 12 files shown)


arendjantetteroo:
> Things to bear in mind before we deploy this and run the import
Overlapping IP ranges now that we have data direct from AWS - We've seen bugs already where the API has been expecting just a single response back, when checking if a domain has an IP inside a given IP address. You can see this documented in issue 48 on the TGWF repo as described below:

> thegreenwebfoundation/thegreenwebfoundation#48

> If this is the case, we'll need to account for this before we merge it, otherwise we might trigger a bunch of 500's.

This only applies to AS numbers as they are globally unique and assigned to one specific entity. There can not be multiple entities with the same AS numbers where the entities are not the same. So if 2 providers have the same AS number they should be considered the same entity. 
Hence the reason why 48 appeared, we had 2 providers that were the same with the same AS but twice in our system. 

> @arendjantetteroo - I know that before running this import, some people who are already customers of Amazon, have listed their IP ranges with us. This means we'd end up with two hosters being listed. How do we handle cases when we have multiple hosters returned?

With regards to ip adress ranges, this is not an issue, there can be multiple providers with ranges that overlap, precisely for this reason. You have a global provider which gives out several of his ip ranges to sub providers, that can have sub hosters etc. 

Our current algorithm looks for the smallest range when returning a result. 
So given
1.1.1.1 - 10.10.10.10 - Global provider A
1.1.1.1 - 3.3.3.3 - Sub provider B
1.1.1.1 - 1.1.1.255 - Local small provider C

When given 1.1.1.200 this will return provider C as a result. 

See also the unit tests in thegreenwebfoundation repo for the greencheck, they contain tests for this logic. 

I've not checked the code (will try to do that later), but this should be fine for our current logic

mrchrisadams:
Linking to the github action that runs this an import/check every day. https://github.com/thegreenwebfoundation/admin-portal/pull/206

Issue #66, opened 2020-03-26
Add richer data for tryout


mrchrisadams:
Right now when you do a check in the internal admin system you get back a basic yes/no answer, like so:

<img width="744" alt="Screenshot 2020-03-26 at 09 44 36" src="https://user-images.githubusercontent.com/17906/77627014-7e904780-6f46-11ea-913d-6d6f017a66e5.png">

This isn't much use for troubleshooting, as it only shows that the API returns.

Ideally we'd show a more detailed view, that also skips any caching we have, the same way that the previous admin did:

![Screenshot_2020-03-26 Greencheck Admin area - tryout - The Green Web Foundation Administration](https://user-images.githubusercontent.com/17906/77627228-da5ad080-6f46-11ea-85d8-06472c9a93a5.png)

### how to add a better tryout

there is a useful library, [ipwhois][1], that provides this kind of info, with a nice API.

Check this usage example:

```
>>>> from ipwhois import IPWhois
>>>> from pprint import pprint

>>>> obj = IPWhois('74.125.225.229')
>>>> results = obj.lookup_rdap(depth=1)
>>>> pprint(results)
```

If we rendered this info, in a template, along with the  current lookup when checking the database, we'd have a a much more useful system.

[1]: https://pypi.org/project/ipwhois/

https://pypi.org/project/ipwhois/

PR #70, opened 2020-04-02
Add more extensive info for trying out urls in admin

mrchrisadams:
This PR addresses #66, introducing similar functionality to the url tryout feature on the older admin view.

### TODO

- [x] show full info returned from API request, including hosting org and last check
- [x] present whois info for a given domain for troubleshooting
- [x] refactor existing form to avoid making multiple requests to the TGWF API on a single page load
- [ ] _maybe_ incorporate green trace 'hops' view when looking up a domain

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -39,6 +39,7 @@ db-to-sqlite = "*"
 sqlite-utils = "*"
 google-cloud-storage = "*"
 sentry-sdk = "*"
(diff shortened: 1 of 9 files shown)


mrchrisadams:
We now have an improved version of this with our `try-out` view, that surfaces ASNs IP ranges and links to similar tooling.

Issue #79, opened 2020-06-03
Allow filtering by dates, and show them in the admin UI


mrchrisadams:
Right now, you can't see when a hoster org was created, and we don't have an easy way of filtering by this. 

Django admin has support for this, and adding will allow users when checking data to focus on a subset of data to verify.

- [x] allow filtering by year of creation
- [x] show the hoster creation date in the listing if we have it

mrchrisadams:
Hi @arendjantetteroo I'm looking over the data in the TGWF database, and I'd like a second opinion.

Based on my understanding of the data, because there is no timestamp for the hosters themselves, my best bet for creating something like this feature for filtering hosting providers for audit, based on the time of their last update would be to look at a corresponding timestamped object, like:

- an IP range approval (the GreencheckIpApprove/ `greencheck_ip_approve` table )
- an ASN approval (via the GreencheckASNapprove / `greencheck_as_approve` table)
- a datacentre (via the HostingproviderDatacenter,  `datacenters_hostingproviders` table)
- a user

I'm thinking if we can see the date of the most recent `created_at`, or `updated_at` date here, it'll give us an idea of which ones we might prioritise for updating.

Can you think of another way, or does this seem sensible at your end?

mrchrisadams:
okay, I've updated this so we have filters by date of last approved:

- IP Range
- ASN
- Datacentre

Two of the queries for this are fast, but one is very slow. I've outlined this in PR #80, and I think it's worth refactoring these in a separate PR, so we can filter this.

PR #80, opened 2020-06-04
First pass at adding date filters in admin

mrchrisadams:
This PR adds date filters for listing hosters, as listed in #79.

### DIFF
--- a/apps/accounts/filters.py
+++ b/apps/accounts/filters.py
@@ -1,6 +1,95 @@
 from django.contrib.admin import SimpleListFilter
 from django_countries.data import COUNTRIES
+from datetime import datetime
+from django.utils import timezone
 
+class YearDCFilter(SimpleListFilter):
+    title = 'Last updated datacentre'
+    parameter_name = 'last_evidence'
+
+    def last_15_years(self):
+        current_year = datetime.now().year
+
+        return ((f"{current_year - x}", f"{current_year - x}") for x in range(15))
+
+    def lookups(self, request, model_admin):
+        return self.last_15_years()
+
+    def queryset(self, request, queryset):
+        if self.value() is None:
+            return queryset
+
+        year = int(self.value())
+        start_at = datetime(year=year, month=1, day=1)
+        end_at = datetime(year=year+1, month=1, day=1)
+
+        for dt in [start_at, end_at]:
+            if timezone.is_naive(dt):
+                timezone.make_aware(dt)
+
+        return queryset.filter(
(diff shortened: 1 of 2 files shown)


mrchrisadams:
Right now, the filter by last IP approval, is pretty damn slow, because the ORM generated SQL query at the mo is something awful like this:

```sql
-- slow query --

SELECT `hostingproviders`.`id`,
       `hostingproviders`.`archived`,
       `hostingproviders`.`countrydomain`,
       `hostingproviders`.`customer`,
       `hostingproviders`.`icon`,
       `hostingproviders`.`iconurl`,
       `hostingproviders`.`model`,
       `hostingproviders`.`naam`,
       `hostingproviders`.`partner`,
       `hostingproviders`.`showonwebsite`,
       `hostingproviders`.`website`,
       COUNT(`greencheck_ip`.`id`) AS `greencheckip__count`
  FROM `hostingproviders`
  LEFT OUTER JOIN `greencheck_ip`
    ON (`hostingproviders`.`id` = `greencheck_ip`.`id_hp`)
 INNER JOIN `greencheck_ip_approve`
    ON (`hostingproviders`.`id` = `greencheck_ip_approve`.`id_hp`)
 INNER JOIN `greencheck_ip_approve` T4
    ON (`hostingproviders`.`id` = T4.`id_hp`)
 WHERE (`greencheck_ip_approve`.`created` > '2006-01-01 00:00:00' AND T4.`created` < '2007-01-01 00:00:00')
 GROUP BY `hostingproviders`.`id`
 ORDER BY `hostingproviders`.`naam` ASC, `hostingproviders`.`id` DESC
```

I don't know MySQL well enough to optimise the SQL query quickly, so I'm posting it here, while I think through alternatives. It's useable, just *slow* at the mo.

apps/accounts/filters.py
```diff
@@ -1,6 +1,95 @@
...
+
+        for dt in [start_at, end_at]:
+            if timezone.is_naive(dt):
+                timezone.make_aware(dt)
+
+        return queryset.filter(
+                greencheckipapprove__created__gt=start_at
+            ).filter(
+                greencheckipapprove__created__lt=end_at
+            )
```
jonathan-s, line 63: Generally you can do
```
queryset..filter(
  greencheckipapprove__created__gt=start_at,
  greencheckipapprove__created__lt=end_at
)
```
Which would mean that it is less than date AND greater than other date. 

There is also the `__range` operator https://docs.djangoproject.com/en/dev/ref/models/querysets/#range

PR #81, opened 2020-06-04
Refactor the year filters

mrchrisadams:
This PR:

- make the classes DRYer, using a mixin instead of inheritance
- uses a range filter query instead of two filters (thanks @jonathan-s  👍 )
- adds tests!

### DIFF
--- a/apps/accounts/filters.py
+++ b/apps/accounts/filters.py
@@ -3,9 +3,7 @@ from django_countries.data import COUNTRIES
 from datetime import datetime
 from django.utils import timezone
 
-class YearDCFilter(SimpleListFilter):
-    title = 'Last updated datacentre'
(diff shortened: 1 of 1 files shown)


jonathan-s:
Performance improved then?

mrchrisadams:
I haven't spent any time trying to speed up the queries yet, but I'm not sure if the filtering is even used yet.

If it is, and it's used often enough to justify the time, then I guess we will but it's less of a priority right now than the other work.

mrchrisadams:
Wow, I've been trying it out, and queries using range turn out to be _much_ faster - as in 20-30x faster now.

Thanks amigo, that's so much better 👍

jonathan-s:
Golden 👌

PR #117, opened 2021-03-06
Resolves #116: Ensure uploaded DB dumps are public

### LINKED ISSUE #116: Add check for updated datasets to be available with public-read permissions
We had a request asking about the datasets at the green_urls endpoint:

https://admin.thegreenwebfoundation.org/green-urls

At present the uploaded objects don't have public read permissions - oops!

### How to fix it:

This [line where we upload to an AWS S3 compatible bucket](https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/greencheck/management/commands/dump_green_domains.py#L49-L50
):

```py
class GreenDomainExporter:
    """
    Exports a snapshot of the green domains, to use for bulk exports
    or making available for analysis and browsing via Datasette.
    """

    # snip  
    def upload_file(self, file_path: str, file_destination: str):
        subprocess.run(["aws", "s3", "cp", file_path, file_destination])

```

We need to explicitly mark this by adding `--acl public-read` to the cli command,, or do it programmatically with boto3.

https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#object

BR0kEN-:
https://github.com/thegreenwebfoundation/admin-portal/issues/116

### DIFF
--- a/apps/greencheck/management/commands/dump_green_domains.py
+++ b/apps/greencheck/management/commands/dump_green_domains.py
@@ -47,7 +47,7 @@ class GreenDomainExporter:
         subprocess.run(["rm", "-rf", file_path])
 
     def upload_file(self, file_path: str, file_destination: str):
-        subprocess.run(["aws", "s3", "cp", file_path, file_destination])
+        subprocess.run(["aws", "s3", "cp", "--acl", "public-read", file_path, file_destination])
 
 
 class Command(BaseCommand):


mrchrisadams:
Awesome, thanks @BR0kEN-!

Issue #119, opened 2021-03-06
Ensure the uploaded dataset is publicly accessible


BR0kEN-:
In #116 we resolved the issue with the uploaded datasets not being publicly available. This issue proposes performing the HTTP HEAD request right after uploading the file and ensuring the response status code is 200.

BR0kEN-:
@mrchrisadams please close this issue since the PR is merged.

PR #120, opened 2021-03-06
#119: Ensure the uploaded dataset is publicly accessible

BR0kEN-:
https://github.com/thegreenwebfoundation/admin-portal/issues/119


mrchrisadams:
Hi @BR0kEN- if you're happy with this to be reviewed, would you tag me mentioning me in a comment or assign using the 'request review' function?

We don't really have a formal convention for requesting code review before merging in yet, so this is a good place as a any to work out one I guess.

I've used a WIP prefix on the PR name previously to indicate its status, but I'm happy to use the platform if it's easier for you.

BR0kEN-:
Hey @mrchrisadams. Yeah, sure, I should have mentioned you. This PR is ready for review.

*The `Request review` is not an option for me though because I don't have enough privileges in this repo.*

BR0kEN-:
Just did a small doc amendment. No more commits to this PR before the review.

apps/greencheck/management/commands/dump_green_domains.py
```diff
@@ -23,62 +36,141 @@ def get_conn_string(self):
...
+        :param file_path: The path to file to compress.
+        :param compression_type: The one of compression types.
+        :returns: The path to created archive.
+        :raises Exception: When `compression_type` is invalid.
+        :raises RuntimeError: When the compression process fails.
+        """
+        if compression_type not in _COMPRESSION_TYPES:
+            raise Exception((
+                f'The "{compression_type}" compression is not supported. Use '
+                'one of "%s".' % ('", "'.join(_COMPRESSION_TYPES.keys()))
```
mrchrisadams, line 75: Would you assign this with a descriptive variable name, and then include it using f-strings like you have elsewhere please? Everything else in this PR is really readable, but I found this harder to scan when reading the code.

apps/greencheck/management/commands/dump_green_domains.py
```diff
@@ -23,62 +36,141 @@ def get_conn_string(self):
...
+        :param db_path: The path to file to export SQLite database to.
+        """
+        cls._subprocess(
+            ["db-to-sqlite", database_url, db_path, f"--table={cls.TABLE}"],
+            f'Failed to export the "{cls.TABLE}" table to "{db_path}" SQLite database.'
         )
 
-    def prepare_for_datasette(self, db_name: str):
+    @classmethod
+    def prepare_for_datasette(cls, db_path: str) -> None:
```
mrchrisadams, line 53: I've realised that this entire method is no longer needed now. It used to make sure that indexes used by sqlite were there, but we now use the same tables to serve requests, so this whole `prepare_for_datasette` method could be delete here, and also in the call to `handle`.

Would you mind removing it this class, and from where it is used in `handle` please

apps/greencheck/management/commands/dump_green_domains.py
```diff
@@ -23,62 +36,141 @@ def get_conn_string(self):
         # change this if we ever leave mysql, obvs
         return f"mysql://{db['USER']}:{db['PASSWORD']}@{db['HOST']}/{db['NAME']}"
 
-    def export_to_sqlite(self, database_url: str, db_name: str):
-        return subprocess.run(
-            ["db-to-sqlite", database_url, db_name, "--table=greendomain"]
+    @classmethod
```
mrchrisadams, line 39: Can you explain the reasoning for using the class and static methods here?

I can understand why you might use them, but then elsewhere in the `Command.handle()` in this management command the object is being instantiated and methods being called as if they regular old methods on an instance of the class.

It might work, but I found this confusing, as I couldn't understand what other value the `staticmethod` and  `@classmethod` decorators added.

I'm totally open to to using class/static methods, but it seem like you'd either want _not_ instantiate objects, and use them, or you'd keep instantiating objects, and then not rely on these decoration of the methods.

I'd prefer we use just one approach, as right now, when I read this I'm unsure if it's better to instantiate, or just call the methods on the class itself.

If I had to make a choice - because there is no real changing state, I think we could get away with not instantiating objects, but it's not something I feel strongly about.

mrchrisadams (changes requested):
Hi @BR0kEN- this looks great! I'm really happy with it, and thank you so much for including the types in places you did. 

There's a couple of small things I need clarification on, but otherwise I'm happy.

One thing though - can you please make sure the files are formatted with black?

I've been meaning to automate formatting with black, but right now it's something I'm doing manually.

apps/greencheck/management/commands/dump_green_domains.py
BR0kEN-: Sure

apps/greencheck/management/commands/dump_green_domains.py
BR0kEN-: Most probably the `GreenDomainExporter` should be refactored because at the moment it is class "on paper" only but, in fact, shares no state among its methods and that's why they're decorated accordingly.

The instantiation (`exporter = GreenDomainExporter()`) is used to perform fewer modifications in the future if the methods become non-static class members.

BR0kEN-:
@mrchrisadams would you mind if I open another PR for adding `black` to the CI process and won't run it for the modified files in the scope of this PR?

mrchrisadams (approved):

Issue #170, opened 2021-06-11
API serving stale cache results


mrchrisadams:
We are holding onto cached results for too long, and this means that end users are seeing out of date results in some cases, especially when a result was grey before, but is green now.

I think this is down to the TTL being too long, and we had the longer TTL, back when making a check was quote an expensive thing to do timewise.

I'm not sure we even need the Redis cache in there any more as most of the time, we are just looking up against a single big-ass table anyway.

### An example

An example would be trefadder.no recently:

When we used the response from the cache it was showing as grey:

https://api.thegreenwebfoundation.org/greencheck/trefadder.no

When we use the non cached version, it was green:

https://api.thegreenwebfoundation.org/greencheck/trefadder.no?nocache=true

### Fixing this

Before we junk the entire cache, a simple way to check this would be to tweak the `nocache` version to it clears out the cache for given domain first before serving the result.

This would mean that just checking for a noncached version would clear the cache, instead of it needing to be cleared manually.

Alternatively, just setting the TTL to be much shorter would do a similar job, as off the top of my head, I'm not sure how long the TTL is on results.

### A bug in the code:

It looks like we have some code which _should_ be doing this already, but for reasons I don't quite understand it's not clearing the cache when we need it to.

https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/greencheck/viewsets.py#L100-L159

I'd welcome some help, as it's almost definitely something _really dumb,_ like checking the wrong Redis database or something similar.

mrchrisadams:
Resolved by #170

PR #171, opened 2021-06-11
Match cache key to invalidate stale data

denning:
Re:  Issue #170

Appears that there was a mismatch between the key get and key delete.

### DIFF
--- a/apps/greencheck/viewsets.py
+++ b/apps/greencheck/viewsets.py
@@ -120,7 +120,7 @@ class GreenDomainViewset(viewsets.ReadOnlyModelViewSet):
 
         # try our cache first before hitting the database
         if not skip_cache:
-            cache_hit = redis_cache.get(f"domains:{domain}")
+            cache_hit = redis_cache.get(f"domain:{domain}")
             if cache_hit:
                 process_log.send(domain)
                 return response.Response(json.loads(cache_hit))


denning:
I wonder if it's just as simple as this? The get/delete names don't match, so I'd expect that would cause some problems here at least.

Could not test due to time constraints - it appears the docs are out of date a bit and and the `docker-compose.yml` file is unmaintained (I'd be happy to bring the docker files up to date, if you think they'd be useful for anyone.)

mrchrisadams:
LOL, it may well be @denning  - I must have been tired when adding that.

I'll try that now, and see if that resolves it.

### Docker and docker compose

Yes, it would indeed be helpful to get the dockerfiles updated. I'll create an issue separately, as I spent a bit of time faffing around with this to get github actions working, and some of that work might by relevant. I'd _really_ like to be able to run this with somethign like gitpod or github codespaces - that would bring down the barriers to entry substantially.

https://github.com/thegreenwebfoundation/admin-portal/blob/master/.github/workflows/run-tests.yml

mrchrisadams:
Oh my. I think that was the issue _was_ the reason. I've checked the naming convention on production and it was indeed missing the `s`. Thanks for this! I'm going to merge it 

https://github.com/thegreenwebfoundation/admin-portal/issues/173

PR #231, opened 2022-04-11
First revision of the documentation.

roald-teunissen:
Initial structure and added installation, deployment, general and how to.

### DIFF
--- a/docs/general.md
+++ b/docs/general.md
@@ -1,5 +1,7 @@
-# General
+# General 
+```{admonition} Draft
 For whom is this documentation intended? x y and z so we will cover ... for a reason.
+```
 
 ## Technologies used
 
@@ -9,6 +11,9 @@ For whom is this documentation intended? x y and z so we will cover ... for a re
 
 ## Project structure
 
+```{admonition} Draft
+The following structure is a placeholder. It can explain the structure of the project with some notable features.
+```
 ```App/
 Controllers/
     User/
@@ -23,21 +28,22 @@ Views/
     Seller/
 ```
 
-## Cheatsheets
-The following list facilitates a usefull set of commands to execute basic operations in this virtual environment.<br> This list serves as a summary of all the commands discussed in the documentation and can be used as guide throughout.  
-
-### General
-|Command   	|Effect   	|
-|---	|---	|
-|```pipenv shell```   	| Enter virtual environment   	|
-|```deactivate```   	| Leave virtual environment   	|
+## Cheatsheets 
+The following list facilitates a useful set of commands to execute basic operations in this virtual environment. This list serves as a summary of all the commands discussed in the documentation and can be used as guide throughout.  
 
 ### Sphinx documentation
 |Command   	|Effect   	|
 |---	|---	|
-|```sphinx-autobuild ./docs _build/```   	| Let Sphinx automatically build the project at every change in the documentation.<br> This operation can be stopped by interrupting the terminal by pressing ```Ctrl + c```    	|
+|`make docs`   	| Build the documentation once.    	|
+|`make docs.watch`  	| Build the documentation and keep automatically updating it until interrupted. This operation can be interrupted by pressing `Ctrl + c`  	|
 
-### Tests
+### Testing
 |Command   	|Effect   	|
 |---	|---	|
-|```make test```   	| Run the test makefile and summarize the results.     	|
+|`make test`  	| Run the test makefile and summarize the results.     	|
+
+### Miscellaneous
+|Command   	|Effect   	|
+|---	|---	| 
+|`pipenv shell`   	| Enter virtual environment(in most cases not necessary as the terminal automatically enters the environment)   	|
+|`deactivate`   	| Leave virtual environment   	|
--- a/docs/general.md
+++ b/docs/general.md
@@ -1,0 +1,43 @@
+# General
+For whom is this documentation intended? x y and z so we will cover ... for a reason.
+
+## Technologies used
+
+|Technology   	|Version   	|
+|---	|---	|
+|       |       |
+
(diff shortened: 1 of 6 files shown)


docs/general.md
```diff
@@ -0,0 +1,43 @@
+# General
+For whom is this documentation intended? x y and z so we will cover ... for a reason.
```
mrchrisadams, line 2: Would you please add an admonition to present this as draft content?

It's ok to include this in the PR but please use an admonition to clearly signpost is as such.

They look like

```
:::{note}
This text is **standard** _Markdown
:::

```

More below

https://myst-parser.readthedocs.io/en/latest/syntax/optional.html?highlight=admonition#admonition-directives

docs/general.md
```diff
@@ -0,0 +1,43 @@
...
+## Cheatsheets
+The following list facilitates a usefull set of commands to execute basic operations in this virtual environment.<br> This list serves as a summary of all the commands discussed in the documentation and can be used as guide throughout.  
+
+### General
+|Command   	|Effect   	|
+|---	|---	|
+|```pipenv shell```   	| Enter virtual environment   	|
+|```deactivate```   	| Leave virtual environment   	|
+
+### Sphinx documentation
```
mrchrisadams, line 35: We spoke about having these as makefile commands. It's cool to have them here, but please also refer to the make command. Ideally, people would use the make commands, leaving us free to change them if need be.

docs/general.md
```diff
@@ -0,0 +1,43 @@
...
+    User/
+    Seller/
+
+Views/
+    User/
+    Seller/
+```
+
+## Cheatsheets
+The following list facilitates a usefull set of commands to execute basic operations in this virtual environment.<br> This list serves as a summary of all the commands discussed in the documentation and can be used as guide throughout.
```
mrchrisadams, line 27: ~~usefull~~ /  useful

I don' think you need the `<br>` either.

docs/index.md
```diff
@@ -13,12 +13,14 @@ Welcome to the documentation for the Green Web platform. This site project conta
 
 This site also provides a sumamry of the other main libraries maintained by the Green Web Foundation, namely `co2js`, a javascript library for understanding and managing the carbon emissions from software, and the green web `grid-intensity` libraries, designed to help design carbon aware digital services that respond to the carbon intensity conditions on the electricity grids the internet relies on.
```
mrchrisadams, line 14: ~~sumamry~~ / summary

mrchrisadams: i think this was _my_ typo from before…

docs/installation.md
```diff
@@ -1,27 +1,30 @@
...
-
-
-
+This installation shows you how to setup the Admin Portal of The Green Web Foundation.<br>
+In order to setup this project, we recommend following the guides from top to bottom. Skipping some topics might result in missing crucial steps for the project to work.
+
+---
+## Gitpod
+### Why Gitpod
+To speed up and simplify the installation process, we use Gitpod. <br> We strongly recommend using this tool.
```
mrchrisadams, line 9: Would you change this to say:

> This is the main supported approach for setting up a development environment. You can use other approaches, but we may not be able to provide as much support.

This based around the idea of a Golden path - there's a blessed approach we take, but people are free to use their own tools as long as they realise we're not able to dedicate the same resources to helping there.

More below.

https://engineering.atspotify.com/2020/08/how-we-use-golden-paths-to-solve-fragmentation-in-our-software-ecosystem/

docs/installation.md
```diff
@@ -1,27 +1,30 @@
...
+
+1. Go to our [Github repository](https://github.com/thegreenwebfoundation/admin-portal) and click on the Gitpod button if you have the browser extention installed or manually create and link it in your Gitpod Workspaces. It might take a couple of minutes to prepare the environment.
+2. After Gitpod is finished preparing the environment, it will open the workspace. Close any unnecessary terminal windows. To open a new fresh terminal window press ```Ctrl + ` ```.
+---
+## Virtual environment
+In an isolated virtual environment we are able to act on various usefull tools within the project.
+
+### Setup
+Before following the following list, make sure you are in the root directory (workspace/admin-portal).
+1. Start by creating a virtual environment named *venv* by executing the following command in a terminal window: ```make venv```.
```
mrchrisadams, line 22: I'm not sure you need three backticks. I think you can just use one at the start, and one at the end.

makefile
```diff
@@ -34,3 +34,9 @@ black.check:
 	@ARGS="--check --color --diff" make black
 
 ci: | black.check flake
+
+docs:
```
mrchrisadams, line 38: Can you please check to see if there is a recommended way to adding a line of documentation to a makefile for a task that might not be obvious?

This might be useful for explaining what each task is supposed to do if the intention isn't immediately obvious.

mrchrisadams (commented):
Hi Roald,

This is good. Only a few changes needed, mainly stylistic and typos.

You're free to merge in without asking for a second review once you have made these changes, but if you want a second review, it's available.

docs/general.md
roald-teunissen: I've added them. I also noticed that this "general" table might get a little cluttered later on as there are a lot of commands that we might see as noteworthy. Perhaps it is either good to keep this as small as possible, have a way of showing the reading that some are important once or just splitting it to other tables. For now I've moved the entering and leaving virtual environment to the miscellaneous table.

makefile
roald-teunissen: At the top of the document you can see that # comments are allowed. I have seen examples where people simply use comments and add echo's to it to show this in the terminal too(https://makefiletutorial.com/). 


In any case, I have added some comments to commands I am almost certain of what they do. Can you expand on the rest?

Issue #189, opened 2021-07-21
Add Azure's list of public ip ranges for green checks.


mrchrisadams:
After reading over the Github Actions documentation I found out that they list their public IP ranges too of the machines running in Azure's various regions.

These are updated each week, so like we do with Amazon, so I think we can do the same for Microsoft's cloud too.

This would mean people building in these regions would get the smiley green badges without loads of tiresome manual updates.

The link to the JSON list of IP ranges and services is below:

https://www.microsoft.com/en-us/download/details.aspx?id=56519

What we do for AWS, we should do for Microsoft too, with a different JSON parser as the data is likely to be structured differently, even if we're still dealing with regions, ASNs and IP ranges.

mrchrisadams:
For reference - here's the management command for AWS. We'd ideally have something like this for Azure, either making a variant of this class, or parameterizing the management command.

https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/greencheck/management/commands/update_aws_ip_ranges.py

And do the same for google too

https://github.com/thegreenwebfoundation/admin-portal/issues/208

Except with GCP and Azure we don't need to distinguish between regions

mrchrisadams:
Linking to the original AWS pull request
https://github.com/thegreenwebfoundation/admin-portal/pull/64

mrchrisadams:
Linking to the github action that runs this an import/check every day. https://github.com/thegreenwebfoundation/admin-portal/pull/206

PR #232, opened 2022-04-19
WIP - build an importer for Microsoft Azure IP ranges

mrchrisadams:
This PR is to resolve issue #189 - building an importer for Microsoft Azure's public IP ranges.

This allows us to link the IP ranges to the organisation, and link to their renewable energy commitments.


apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+from apps.greencheck.models import GreencheckIp
+from apps.greencheck.management.commands.update_azure_ip_ranges import GREEN_REGIONS
+
+from django.conf import settings
+
+
+@pytest.fixture
+def hosting_provider():
+
+    # oregon, *rest = [region for region in GREEN_REGIONS if region[1] == "us-west-2"]
```
mrchrisadams, line 19: Would you please remove this part? It's for the other provider

apps/greencheck/management/commands/update_azure_ip_ranges.py
```diff
@@ -0,0 +1,104 @@
...
+
+logger = logging.getLogger(__name__)
+
+GREEN_REGIONS = (
+    ("Azure", "azure-code", settings.AZURE_PROVIDER_ID),
+)
+
+class MicrosoftCloudProvider:
+    def retrieve(self):
+        with open('apps/greencheck/management/commands/azure-ip-ranges.json') as json_file:
```
mrchrisadams, line 19: Rather than using this string, would you please use the Pathlib for locating this file? It makes it cross platform, and the API is fairly nice to use. This documentation here shows some sample uses to copy, and explains why you might use it in more detail.

https://treyhunner.com/2019/01/no-really-pathlib-is-great/

mrchrisadams, line 19: Rather than using this string, would you please use the Pathlib for locating this file? It makes it cross platform, and the API is fairly nice to use. This documentation here shows some sample uses to copy, and explains why you might use it in more detail.

https://treyhunner.com/2019/01/no-really-pathlib-is-great/

apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+        model="groeneenergie",
+        name=name,
+        partner="",
+        showonwebsite=True,
+        website="http://azure.microsoft.com",
+    )
+
+
+@pytest.fixture
+def azure_cloud_provider(hosting_provider):
```
mrchrisadams, line 37: Would you please add a doc string outlining what this is returning and why?

I know right now, but I think we'll forget in a month's time why were returning the instantiated cloud provider rather than the hosting provider instead.

mrchrisadams, line 37: Would you please add a doc string outlining what this is returning and why?

I know right now, but I think we'll forget in a month's time why were returning the instantiated cloud provider rather than the hosting provider instead.

apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+    hosting_provider.save()
+    return update_azure_ip_ranges.MicrosoftCloudProvider()
+
+
+@pytest.fixture
+def azure_json_ip_ranges():
+    this_file = pathlib.Path(__file__)
```
mrchrisadams, line 44: Nice! Thanks you for using pathlib here! 💚  this is a good norm to apply for all access of local files to read

mrchrisadams, line 44: Nice! Thanks you for using pathlib here! 💚  this is a good norm to apply for all access of local files to read

apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_fetch_ip(self, hosting_provider, azure_cloud_provider):
+        """
+        Do we fetch the data from Azure?
+        """
+        hosting_provider.save()
+        res = azure_cloud_provider.retrieve()
+
+        assert len(res.keys()) == 3 # TODO: Figure out what value this should be
+
+    def pullout_green_regions(self, hosting_provider, azure_cloud_provider):
```
mrchrisadams, line 62: would you please add a docstring for why this is necessary, and what it returns?

This guidance is helpful for documenting code for python
https://realpython.com/documenting-python-code/#commenting-code-via-type-hinting-python-35

mrchrisadams, line 62: would you please add a docstring for why this is necessary, and what it returns?

This guidance is helpful for documenting code for python
https://realpython.com/documenting-python-code/#commenting-code-via-type-hinting-python-35

apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+
+        _, region, host_id = azure_cloud_provider.green_regions[0]
+
+        iprs = azure_cloud_provider.pullout_green_regions(res, region)
+
+        ip_ranges = azure_cloud_provider.ip_ranges_for_hoster(iprs)
+
+        assert isinstance(ip_ranges[0], ipaddress.IPv4Network)
+
+    def test_update_hoster(self, hosting_provider, azure_cloud_provider):
```
mrchrisadams, line 73: Would you please add a docstring explaining why this test is helpful? i.e explain why we check that we can update the hoster in this case, and what role the azure cloud provider plays here in this test. If this fails, what are the consequences of it failing for the system?

mrchrisadams, line 73: Would you please add a docstring explaining why this test is helpful? i.e explain why we check that we can update the hoster in this case, and what role the azure cloud provider plays here in this test. If this fails, what are the consequences of it failing for the system?

apps/greencheck/tests/test_azure_cloud_importer.py
```diff
@@ -0,0 +1,114 @@
...
+        assert GreencheckIp.objects.all().count() == 0
+
+        hosting_provider.save()
+
+        # create one new green IP range
+        azure_cloud_provider.update_hoster(hosting_provider, ip_start, ip_end)
+
+        assert GreencheckIp.objects.all().count() == 1
+
+    def test_process(
```
mrchrisadams, line 89: would you please add a longer method name here, and/or add a docstring about what process we're testing? It wasn't obvious to me, and it's one of the first places we'd look when trying to understand what it would mean if this test was failing

mrchrisadams, line 89: would you please add a longer method name here, and/or add a docstring about what process we're testing? It wasn't obvious to me, and it's one of the first places we'd look when trying to understand what it would mean if this test was failing

docs/how-to.md
```diff
@@ -0,0 +1,4 @@
+# How to use..
```
mrchrisadams, line 1: I think this branch might need a rebase to include the updated documentation if it's already in master, or have the other PR merged in then rebased here.

I've approved the earlier PR to merge in, and just clicked to merge it in. You might need to try rebasing this branch to include the updates docs from it.

The link below shows the link to the me approving your earlier documentation work, and I mentioned you're free to merge it in, but it might not have been obvious.

https://github.com/thegreenwebfoundation/admin-portal/pull/231#pullrequestreview-942040899

mrchrisadams, line 1: I think this branch might need a rebase to include the updated documentation if it's already in master, or have the other PR merged in then rebased here.

I've approved the earlier PR to merge in - would you please merge it in, then try rebasing this branch to include the updates docs from it.

https://github.com/thegreenwebfoundation/admin-portal/pull/231#pullrequestreview-942040899

If it isn't this would probably be better as a separate PR in future.

mrchrisadams (commented):
Hi @roald-teunissen this is looking good. I've requested a few updates of documentation mainly, and I think we're better off pulling in the json file from a separate object storage bucket rather than keeping it in the code base.

Can you please let me know if you are able to put it in a bucket we can fetch it from instead?

roald-teunissen:
> Hi @roald-teunissen this is looking good. I've requested a few updates of documentation mainly, and I think we're better off pulling in the json file from a separate object storage bucket rather than keeping it in the code base.
> 
> Can you please let me know if you are able to put it in a bucket we can fetch it from instead?

It would be better if Microsoft shared their endpoint with us. I've seen a couple of endpoints going around on stackoverflow, but those are mainly dynamic and time-bound. Can you perhaps ask to get access for access to one of their endpoints?

On the other hand, if we plan on doing it ourselfs we need to download the json and put it in object storage, which seems like a unnecessarily complicated workaround. If I see any endpoints that are reliable, I'll let you know.

apps/greencheck/tests/test_azure_cloud_importer.py
roald-teunissen: It has been removed.

As there is only one region, I've also left that part out. 
Do you think we should name this service "Azure" or "Microsoft", or something else?

roald-teunissen: Also, for now I have used 123 as hosting id. Can this be any arbitrary number?

apps/greencheck/tests/test_azure_cloud_importer.py
roald-teunissen: That's a good question. I've looked at the code of the AWS example and we also check the number of keys there.
I am guessing we are checking the general structure of the json. 

Why did we use this in the AWS code? I can't really find a reason behind doing this in the first place

mrchrisadams:
Hi Roald,

Thanks for the reply:

> As of writing this, the following link can be retrieved: https://download.microsoft.com/download/7/1/D/71D86715-5596-4529-9B13-DA13A5DE5B63/ServiceTags_Public_20220425.json
> 
> This link might change in the future, so this is not reliable. But if we can figure out what this represents, we might be able to use it. And yes, using the last link works. I have tested it. So if we can use this, we don't need object storage.

I'm reading this `20220425` fragment of the file to mean 2022-04-25 as a sort of timestamp for the file.

One thing we could do is make the update command that fetches the json fetch it from a provided url defined in a config file, then safely abort the import if fetching it isn't successful.

This would give us a graceful way to run this idempotently, as well as in testing, and as long as the json file has the same kind of data in it, we could even keep it up to date by changing this URL used on a regular cron job

roald-teunissen:
In addition to the documentation changes, it would be nice to have type hinting.

roald-teunissen:
> Hi Roald,
> 
> Thanks for the reply:
> 
> > As of writing this, the following link can be retrieved: https://download.microsoft.com/download/7/1/D/71D86715-5596-4529-9B13-DA13A5DE5B63/ServiceTags_Public_20220425.json
> > This link might change in the future, so this is not reliable. But if we can figure out what this represents, we might be able to use it. And yes, using the last link works. I have tested it. So if we can use this, we don't need object storage.
> 
> I'm reading this `20220425` fragment of the file to mean 2022-04-25 as a sort of timestamp for the file.
> 
> One thing we could do is make the update command that fetches the json fetch it from a provided url defined in a config file, then safely abort the import if fetching it isn't successful.
> 
> This would give us a graceful way to run this idempotently, as well as in testing, and as long as the json file has the same kind of data in it, we could even keep it up to date by changing this URL used on a regular cron job

That sounds like a great solution. I was thinking the same although I'm not sure about the 7/1 and the numbers following that. I will keep an eye open for this endpoint and see if it changes overtime. It might be as you say it is. Otherwise we might need some help deciphering this.

mrchrisadams:
OK, I've added a few small updates for this - mainly small stylistic changes while I was testing this out locally. I'm merging this in now. Thanks for all your work on this, @roald-teunissen !

PR #240, opened 2022-05-05
Ca 190 equinix importer - resolves #190

### LINKED ISSUE #190: Add Equinix's datacentre into the listing
Equinix are a big player and they release data on a regular basis about how their datacentres are powered, but it's a PDF.

We should find a way to load this into the system, although doing it manually would be a chore

https://sustainability.equinix.com/wp-content/uploads/2020/11/GU_IBX-Sustainability-Quick-Reference_US-EN-4.pdf

Their DC map might have the same data in json form

roald-teunissen:
An importer that extracts IPv4, IPv6 and Autonomous System Numbers (ASN) for all of Equinix's datacentres, and updates the entries in the database.


mrchrisadams:
Thanks for this @roald-teunissen. This is looking good, and I think once we run this import, we can close #190 below:

https://github.com/thegreenwebfoundation/admin-portal/issues/190

However, I see some issues related to being able to rebase it. you might need to rebase this branch from master now that we've merged in the Microsoft Importer.

Issue #260, opened 2022-06-24
[BUG] Update inactive domains when importing


roald-teunissen:
**Describe the bug**
Domains are being shown as active even though the domain hosting provider does not include them in their list of active domains.

**Expected behavior**
When importing the dataset containing all active domains, it is expected that domains that are not in the list are being set to inactive.

**Actual behavior**
When importing new domains from the dataset of the hosting provider, it simply updates domains to active and doesn't take action on the other domains.

roald-teunissen:
![](https://github.trello.services/images/mini-trello-icon.png) [Add cache busting feature when IP ranges are marked as inactive](https://trello.com/c/Svc9RBTU/35-add-cache-busting-feature-when-ip-ranges-are-marked-as-inactive)

PR #264, opened 2022-07-01
Update inactive domains when importing

roald-teunissen:
Records were only set to active and never back to inactive if they did not appear in the list given by the hosting provider.
In addition to this change I also improved the following:
- Database atomicity. Making it possible to roll back a database transaction if it goes wrong.
- More comprehensive testing

[See this issue for more info](https://github.com/thegreenwebfoundation/admin-portal/issues/260).

### DIFF
--- a/apps/greencheck/importers/importer_interface.py
+++ b/apps/greencheck/importers/importer_interface.py
@@ -1,7 +1,6 @@
 import ipaddress
 import re
 import logging
-import pdb
 
 from typing import Protocol, runtime_checkable, List
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


apps/greencheck/importers/importer_interface.py
```diff
@@ -17,67 +19,138 @@ def fetch_data_from_source(cls) -> list:
...
+                        active=True, asn=asn_network, hostingprovider=hosting_provider
+                    )
+                    entry.save()
+
+                    if created:
+                        logger.debug(entry)
+
+        return updated_networks
+
+    def update_ip(cls, active_networks: List[str]) -> int:
```
mrchrisadams, line 105: Please add a docstring like the other methods 👍

apps/greencheck/importers/importer_interface.py
```diff
@@ -17,67 +19,138 @@ def fetch_data_from_source(cls) -> list:
...
+                if (entry.ip_start, entry.ip_end) in active_networks:
+                    active_networks.pop(
+                        active_networks.index((entry.ip_start, entry.ip_end))
+                    )
+                    entry.active = True
+                    entry.save()
+
+                    updated_networks += 1
+                    logger.debug(entry)
+                elif entry.active == True:
```
mrchrisadams, line 135: Can you add a one liner comment about why we make things as true or false, or add it at the top of the loop explaining the intent here?

mrchrisadams (commented):
Hi @roald-teunissen thanks for this. I'm largely happy with the implementation, but it would really help to have the docstrings filled out and a  bit more documentation about the implemetation.

Once you have made those changes feel free to merge it in, as I'm happy with what the code does, but in it current state I worry about us not remembering what it does in a few months and wishing we had the extra documentation 😅

roald-teunissen:
Thanks @mrchrisadams for the insightful comments.
I've taken your feedback to heart and even found myself being confused after reading some of old code after so long.

Hopefully the comments I made in today's commit might help the reader to understand the code a bit better. 
Especially this interface needs to be pristine as it holds core functionalities for our other importers that extend it.

mrchrisadams:
Recent changes in #532, as prep for some later work has superceded this PR. There are some good ideas here we might want to come back to  and borrow though, like atomic transactions when running an automated imports for example.

Issue #266, opened 2022-07-06
[BUG] Company names not in alphabetic order


roald-teunissen:
**Describe the bug**
Hosting companies displayed on [the website](https://www.thegreenwebfoundation.org/directory/) and also retrieved from the API are not in alphabetic order.

**To Reproduce**
1. Retrieve hosting provider companies through the API as JSON or go to [the website](https://www.thegreenwebfoundation.org/directory/)

**Expected behavior**
Company names underneath each country is in alphabetic order.

**Actual behavior**
Company names are not in alphabetic order

**Visual proof**
![image](https://user-images.githubusercontent.com/8720223/177537074-95a41c92-668c-402b-b964-2782627b9e2d.png)

**Additional context**
[Check the Trello card that is connected to this issue.](https://trello.com/c/JZ6rTmcg/166-support-request-from-r-tuk-alphabetical-order)

roald-teunissen:
![](https://github.trello.services/images/mini-trello-icon.png) [Support request from: R Tuk (alphabetical order)](https://trello.com/c/JZ6rTmcg/166-support-request-from-r-tuk-alphabetical-order)

tortila:
Just to confirm if I understood the issue: the expected behaviour is that the providers are sorted by: partner first, then alphabetically by name?

PR #267, opened 2022-07-06
Implemented ordering functionality for providers in directories

roald-teunissen:
[Check this issue.](https://github.com/thegreenwebfoundation/admin-portal/issues/266)


roald-teunissen:
Hi @mrchrisadams, thanks for improving upon this!
I'm not really understanding why we want to include the partners and non-partners? 
In this context do we refer to partners as green hosting providers? If so, we only want to return a list of countries accompanied with their green hosting providers (partners) right?

PR #300, opened 2022-09-07
Add extensions for python development in gitpod

mrchrisadams:
This PR introduces the use of pre-installed VS code extensions, so we automatically can benefit from:

- using black for formatting by default
- flake8 for catching common linting issues, rather than needing to bring them up in code review
- isort for easier to read module import orders

More below:

https://www.gitpod.io/docs/ides-and-editors/vscode-extensions

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -32,7 +32,7 @@ tasks:
       npx rollup --config
       cd ../../../
       pipenv run ./manage.py collectstatic --no-input
-      
+
     command: ls
 
   - name: rabbitmq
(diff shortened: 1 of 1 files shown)


mrchrisadams:
Gitpod has a number of extensions you can add for python that look very useful but have not been added. See below:

https://www.gitpod.io/docs/languages/python/

If you are not working in a browser you may need to update your desktop version of VS code to use the Gitpod settings sync over the one provided by Microsoft or Github. This will ensure settings you are used to locally are stored so you can access them when you spin up a dev instance hosted by Gitpod.

More below:
https://www.gitpod.io/docs/ides-and-editors/settings-sync#enabling-settings-sync-in-vs-code-desktop

Issue #292, opened 2022-08-31
[BUG] A check within an registered IP range is not showing as green


mrchrisadams:
**Describe the bug**

From a support discussion:

> Regarding the Green Web Checker, I checked your website myself, and it is indeed turning up grey. This is strange, because the underlying IP address (103.209.24.129) is included in the IP ranges that were approved as green.

This start happening in the last day or so. I'm not sure why.

**To Reproduce**

Check the url using the 'debug' checker

https://admin.thegreenwebfoundation.org/admin/try_out/

**Expected behavior**

It should resolve to a green result.

**Actual behavior**
_A clear and concise description of what actually happened._

We see a grey result.

**Additional context**

We have developer facing documentation explaining how our green check system works.

https://greenweb.readthedocs.io/en/latest/how-we-check-for-green.html

mrchrisadams:
![](https://github.trello.services/images/mini-trello-icon.png) [Serversaurus IPs not working?](https://trello.com/c/vIrAHcB6/281-serversaurus-ips-not-working)

mrchrisadams:
OK, this was actually a typo in the end, where the IP range had been entered incorrectly through the admin, but it it did highlight a need for validation.

@tortila and I have spoken about this, and we it makes sense to add the validation at the GreencheckIP model linked below:

https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/greencheck/models/checks.py#L153

Adding model-level validation ought to cascade up through the form submission lifecycle, and show up in the admin, but _also_ give us a way to use the same model level checking for any other import we do, like via the API, or through a CSV import.

tortila:
The PR I linked to this issue solves the IP range validation for `ModelForm`s, e.g. used in the Django admin. There are possibly other places where this model is saved without checking if it's valid, e.g. CSV import: https://github.com/thegreenwebfoundation/admin-portal/blob/bff33eb980b74073d065e47861aff05962996d25/apps/greencheck/bulk_importers.py#L54-L57

I also found out that for the REST API, the serializer defined for `GreencheckIp` model already has similar validation logic implemented: https://github.com/thegreenwebfoundation/admin-portal/blob/9d261aaa36b86aa6d786aa92203f429576190d19/apps/greencheck/serializers.py#L73-L84.

PR #301, opened 2022-09-07
Validate IP ranges

tortila:
Closes https://github.com/thegreenwebfoundation/admin-portal/issues/292.

Scope of changes:
- Added model-level validation to `GreencheckIp` and `GreencheckIpApprove` models. Now newly submitted IP ranges (via `ModelForm`s that are using those models) will be automatically validated - tested manually in Django admin
- Added a test for the validation logic
- Changed the order of definition of fields in `GreencheckIp` and `GreencheckIpApprove` models. Thanks to this small change `IP start` is now displayed before `IP end` in the admin panel (hopefully making it more intuitive to input the data)
- Fixed some typos and formatting (in the files that I worked on)

### DIFF
--- a/apps/greencheck/admin.py
+++ b/apps/greencheck/admin.py
@@ -8,7 +8,7 @@ from apps.accounts.utils import reverse_admin_name
 
 from . import forms, models
 from .choices import StatusApproval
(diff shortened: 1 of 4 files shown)


mrchrisadams (approved):
This is great - I'm really happy with this. Thanks @tortila !

PR #302, opened 2022-09-09
Add missing migration

tortila:
There seems to be a mismatch between the state of the Django models and the database - Django complains about it every time I start the server. I followed the suggestion and generated the missing migrations:
```
$ pipenv run ./manage.py makemigrations
```
It looks like some of the changes are messing with Enum fields of the biggest table: `greencheck`, I find it difficult to judge how safe it is to run those migrations.

Another point is that when I apply this migration locally, it fails:
```
$ pipenv run ./manage.py migrate
[...]
django.db.utils.OperationalError: (1051, "Unknown table 'greencheck.greencheck_stats'")
```
I'm curious if this behavior is the same on other developers' machines?

In this specific case I verified that indeed, my local database does not have such table. So maybe we will have to manually adjust the migrations that were automatically generated by Django. 

I suggest we keep this PR as a point for discussion about how to resolve this inconsistent state. This is not urgent but can quickly become that if we have other changes in models that will require running migrations.


mrchrisadams:
@tortila I'm removing this as we're doing an audit of the tables we use anyway in the coming weeks which would likely take precedence over this work. you can see the issue below for adding any notes in the public domain 

https://github.com/thegreenwebfoundation/admin-portal/issues/498

PR #307, opened 2022-09-21
253 microsoft import endpoint refresher

roald-teunissen:
Work in progress! 
Missing a working access token to test the bucket's functionality.
Can you please schedule a meeting with me to discuss important design decisions that need to be made?

This change includes the following:
- Object storage refactoring and additions
- - New put and get functions for specific files in object storage
- - Ability to retrieve a specific bucket
- - Ability to retrieve a resource session for additional actions
- Microsoft's dataset refresher
- - Searcher for checking workable endpoints from Azure source
- - Updater to overwrite an older version of the dataset that has been saved in our bucket


apps/greencheck/tests/test_microsoft_bucket_updater.py
```diff
@@ -0,0 +1,65 @@
...
+        Test the searching function to track an active endpoint
+        """
+        updater = MicrosoftBucketUpdater()
+
+        # Use dataset searcher
+        dataset = updater.search_dataset()
+
+        assert len(dataset) > 0
+
+    def test_update_bucket(self):
```
mrchrisadams, line 23: Yeah, this is going to be a bit tricky to test, as we're relying on an external API on a domain we don't own, and the presence of files there to tell us whether we make updates.

Previously I've used the `mocker` class to mock out requests to external services and in this case I think we might need to as well.

I can show you on the call this morning

apps/greencheck/microsoft_bucket_updater.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        date_to_check = datetime.now()
+        date_threshold = datetime.now() - timedelta(days=5)
+
+        # Iterate through the days to find a working
+        # endpoint that returns a dataset
+        while date_to_check >= date_threshold:
+            response = requests.get(self.format_date_to_url(date_to_check))
+
+            if response.status_code == 200:
```
mrchrisadams, line 31: See above. returning just the URL would mean this method is only telling us if a url is good to use, not returning any content. We already have methods for fetching the code,  and personally, I'd prefer less complex, more composable code over saving on making HTTP requests.

mrchrisadams (changes requested):
Hi @roald-teunissen  - I've made a few comments. This is largely going in the right direction, but it'll be good to discuss this on the call, as you've raised a few questions about how best to implement this.

mrchrisadams:
Hey @roald-teunissen  I figured it might be useful for me to share how I imagined this would work, in advance of the call:

1. Fetch the url defined at `MICROSOFT_IP_RANGES_ENDPOINT`.
2. Parse out the date from the url. Return early if the date is the same as today, doing nothing.
4. If the date is less than today, start looking for a new json file to import from our source., by either searching forwards from the parsed date, or backwards from the current date until you reach it. ^1
5. If no new files are found, do nothing, and return early.
6. If a new file is found at a greater date, then upload it to MICROSOFT_IP_RANGES_ENDPOINT, so the importer can use on future updates, then return. 

If we do it this way, it will mean the daily import can consume the the file at the new URL, and decouples the _importing_ of new IP range files, from the _finding_ of new IP range files


^1: In both cases the number should be the same, but the backwards search might go on forever if no valid url is ever found in object storage. We should have valdi urls though.

apps/greencheck/tests/test_microsoft_bucket_updater.py
```diff
@@ -0,0 +1,65 @@
...
+
+        assert date == updater_date
+
+    def test_format_date_to_url(self):
+        """
+        Test formatting from date to url
+        """
+        updater = MicrosoftBucketUpdater()
+        date = datetime.now()
+        url = updater.url_prefix + date.strftime("%Y%m%d") + updater.url_file_extension
```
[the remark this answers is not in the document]
roald-teunissen: Aha thanks! I was looking for this!

apps/greencheck/microsoft_bucket_updater.py
```diff
@@ -0,0 +1,63 @@
...
+    def search_dataset(self) -> dict:
+        """
+        Searches for a working url from Microsoft's endpoint.
+        If a dataset is found, it will be returned by this function.
+        Return: dict
+        """
+        date_to_check = datetime.now()
+        date_threshold = datetime.now() - timedelta(days=5)
+
+        # Iterate through the days to find a working
```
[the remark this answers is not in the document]
roald-teunissen: Do you maybe have a suggestion for a better name for this function?
I was thinking of `` search_endpoints_for_dataset`` for example

.env.gitpod
```diff
@@ -27,7 +27,7 @@ MAXMIND_USER_ID = 123456
 MAXMIND_LICENCE_KEY = "xxxxxxxxxxxxxxxx"
 
 MICROSOFT_PROVIDER_ID = 1234
-MICROSOFT_LOCAL_FILE_DIRECTORY = "https://tgwf-web-app-test.s3.nl-ams.scw.cloud/data-imports/ms-azure-ip-ranges-2022-04-25.json"
+MICROSOFT_REMOTE_FILE_DIRECTORY = "https://tgwf-web-app-test.s3.nl-ams.scw.cloud/data-imports/ms-azure-ip-ranges-2022-04-25.json"
```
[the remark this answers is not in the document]
roald-teunissen: Thanks for noticing.
Honestly, I'm not sure what happened here. It does make sense to name it the same way as the others. 
Changed it right away.

roald-teunissen:
Hi @mrchrisadams, I've processed your feedback and tested my code against my newly created object storage as you suggested.

I have implemented the following:
**Updater**
- Comprehensive docstrings, comments and improved variable names
- Updated Microsoft's API endpoint constant
- Searching backwards in time is done from today until the last updated date, extracted from the name of the file in object storage
- Improved formatting functions and fixed an error

Please ignore the test file for now. There is still  a lot to be done there before we can ship it. 

For now, can you please review the following:
- [ ] microsoft_bucket_updater.py
- [ ] object_storage.py
- [ ] additional comments

apps/greencheck/object_storage.py
```diff
@@ -3,30 +3,53 @@
...
 """
 import boto3
+import json
 from django.conf import settings
 
 
-def object_storage_bucket(bucket_name: str):
+# TODO: how to specify this return time. The following example throws an error as
+# this structure does not exist
+# def retrieve_object_storage() -> boto3.resources.factory.s3.ServiceResource:
```
roald-teunissen, line 12: I can't seem to specify the return time like this.
This is what the documentation of boto3 and simply type() point out, but it is not accepted because the problem seems to be that s3 is not an attribute of factory.

.env.gitpod
```diff
@@ -27,7 +27,7 @@ MAXMIND_USER_ID = 123456
 MAXMIND_LICENCE_KEY = "xxxxxxxxxxxxxxxx"
 
 MICROSOFT_PROVIDER_ID = 1234
-MICROSOFT_LOCAL_FILE_DIRECTORY = "https://tgwf-web-app-test.s3.nl-ams.scw.cloud/data-imports/ms-azure-ip-ranges-2022-04-25.json"
+MICROSOFT_REMOTE_API_ENDPOINT = "https://tgwf-web-app-test.s3.nl-ams.scw.cloud/data-imports/ms-azure-ip-ranges-2022-04-25.json"
```
roald-teunissen, line 30: The endpoint seems to specify a date at the end. My code is reliant on that date as it extracts it and calculates the days between today and last updated. Is there a reason it is a literal?

roald-teunissen, line 30: What do you think of renaming all these constants from:
`PROVIDER_REMOTE_API_ENDPOINT`
to
`PROVIDER_OS_ENDPOINT` or `PROVIDER_TGWF_OS_ENDPOINT`

roald-teunissen, line 30: We seatled with this structure because we made requests to this url in the importer (the constant is only used once).
We can improve this by using the `object_storage` class and calling the bucket from there. We would only need to set this constant as:
`MICROSOFT_REMOTE_API_ENDPOINT = "data-imports/ms-azure-ip-ranges-"`
(As I suggested on one of my other comments, the naming could be improved)

With this prefix we can do a filter function on the bucket to get the latest version of that file, which already dynamically changes its name based on when it is lastly updated.
By doing this, the constant can easily be used by more files as this prefix is also needed elsewhere.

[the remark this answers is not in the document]
mrchrisadams: Would you mind if we kept this? I'll try laying out my reasoning, which was largely `{company}-{provider}-{content}-{date}`, while trying to avoiding having really long variable names:

- microsoft is the _company_. `ms` is a common abbreviation, just like Amazon might is sometimes shortened to `AMZN`, or google `GOOG`. I accept that using `microsoft` might be clearer if you're coming new to it.
- `azure` is the brand Microsoft uses as a provider of hosted cloud services. Just like `aws` is Amazon's brand Amazon Web Services, and `gcp` is Google Cloud Platform the brand from Google/Alphabet. This is why initially the environment variables for the AWS importer used AWS and not Amazon. One is a service from the org, even if Amazon is the _reporting entity_ that publishes sustainability reports.
- we only have `ip-ranges` in use here, not AS networks or any other formulation. So, there is only one way of describing networks inside the data we get from the endpoint. You could arguably use `network-ranges`, as a convention for all bulk import
- the 2022-04-25 format is an ISO format ([ISO 8601]()https://www.iso.org/iso-8601-date-and-time-format.html), specifically designed to remove ambiguity, and be easy to parse. When you remove the spaces, it's much harder to tell if you're following a YYYY-MM-DD format. When you lose the spaces, and just have a string of numbers it becomes ambiguous.  When you don't have was a figures like 20220425. This could be interpreted as _20:22 on the 4th of April_, or if we were working with an earlier year of data like 2012, 20120425 could easily be interpreted as 20th December (i.e. month 12) at 04:25.

mrchrisadams:
hey @roald-teunissen  I won't be able to get to this until Friday.

I'll be able to give this a proper code review on Friday

apps/greencheck/microsoft_bucket_updater.py
```diff
@@ -0,0 +1,116 @@
...
+        latest version of the dataset first, in the case of having multiple endpoints
+        active at once.
+
+        Return: dict (representing a dataset from Microsoft containing active IP ranges)
+        """
+        # Set a range to start searching from:
+        # - pivot : iterable date, starting with today
+        # - until : date to search up to (difference between today and last updated)
+        date_pivot = datetime.now()
+        duration_between_last_updated = self.retrieve_delta_last_updated()
```
roald-teunissen, line 32: @mrchrisadams we've talked about this on our call today.
We talked about searching for a endpoint that returns a dataset from today until a certain point.

Firstly we discussed that we can search all the way back to the first ever updated. Secondly we discussed that it was simply one week. But it seems to be easier than that. As you can see, the function _retrieve_delta_last_update_ simply retrieves the time between now and the date of the last updated file in our object storage. This makes it much more reliable and can therefore dynamically adjust in relation to Microsoft's updating schedule.

For example, if the last updated was a week ago, then it will search for 7 days. Simply because there are 7 days between today and the one that was last updated. If this was one month ago, it will search the whole month until it finds a working endpoint. If it finds something, then it will add it to the object storage and that will be our "last updated" file.

roald-teunissen:
Need to find out: does the dataset contain **only IP ranges** or does it also contain **ASN**

mrchrisadams:
I'm closing this now since we refactored the importers/

PR #318, opened 2022-10-13
Update how we show datacentres - fixes #317

### LINKED ISSUE #317: [ENHANCEMENT]  Restrict association of datacentres with providers to green web staff members
**Is your feature request related to a problem? Please describe.**

Right now, if you are a hosting provider, and you want to show that you are using a specific datacentre, you need to visit the datacentre page page, and then list your provider. 

Doing so triggers a notification for a staff member to review this information, and make sure it's correct, but it's error prone and we have had at least one case where we have had a provider list incorrect information.

We currently link the information for deciding whether to show a green result versus a grey result based on the IP network information that a hosting provider has shared with us (ie. their IP ranges or AS numbers), so this link does not materially change the status of a check.

However, it is confusing, when we list this in the directory.

We are also updating our data model to remove the notion of a datacentre in favour of having  supply chain of providers who provide 'colocation' as one of a number of services. 

This means that the idea of datacentre as first class objects in our admin interface makes less and less sense over time.

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._

To avoid more confusion, it makes sense to remove the hosting provider picker from the datacentre  screen for non-staff users, and instead provide a link to our support form. This should allow existing datacentre operators to request changes for our support staff until we have the new data model available to end users.

It will also stop end users accidentally linking their provider to the wrong datacentre and vice versa.

mrchrisadams:
This PR changes the how we present the data centre page to only allow green web staff to associate providers with a given datacentre.

Previously we allowed hosting providers to declare this connection, which we would validate afterwards, but this switches it around so that we only update information once we have had a request to update it, and approved the change.

**What normal, non staff users see**

With this PR merged in, regular users see something like this example provider. You can see hosting providers associated with the datacentre, but you can not change them. We link to a support form to request changes. Very few changes are made to datacentres each month (I only saw 6 in Sept) so this doesn't sound like it would create a massive support burden.

Screenshot 2022-10-13 at 12 17 35

**What green web staff see**

Our internal admin staff see a similar view, but have the widget to make a change too now.

Screenshot 2022-10-13 at 12 17 48

### DIFF
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -960,14 +960,12 @@ class DatacenterAdmin(admin.ModelAdmin):
             datacentre = self.model.objects.get(id=object_id)
             associated_providers_count = datacentre.hostingproviders.all().count()
 
-            if associated_providers_count > 1000:
-                extra_context["associated_providers_count"] = associated_providers_count
-                extra_context["associated_providers"] = []
-            else:
+            if associated_providers_count:
                 extra_context["associated_providers_count"] = associated_providers_count
                 extra_context[
                     "associated_providers"
                 ] = datacentre.hostingproviders.all()
+                extra_context["dc_has_providers"] = True
 
         return super().change_view(
             request,
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -149,7 +149,9 @@ class CustomUserAdmin(UserAdmin):
         # what we show for internal staff
         staff_fieldsets = (
             "Permissions",
-            {"fields": ("is_active", "is_staff", "groups"),},
+            {
+                "fields": ("is_active", "is_staff", "groups"),
+            },
         )
 
         # our usual set of forms to show for users
@@ -202,8 +204,7 @@ class HostingProviderSupportingDocumentInline(admin.StackedInline):
 
 
 class HostingProviderNoteInline(admin.StackedInline):
-    """
-    """
+    """ """
 
     extra = 1
     model = HostingProviderNote
@@ -323,10 +324,8 @@ class HostingAdmin(admin.ModelAdmin):
             messages.add_message(
                 request,
                 messages.WARNING,
-                (
-                    "No user exists for this host, so you will need to "
-                    "add an email manually"
-                ),
+                "No user exists for this host, so you will need to "
+                "add an email manually",
             )
 
         context = dj_template.Context({"host": obj, "recipient": user})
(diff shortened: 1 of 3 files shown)


mrchrisadams:
I've seen a failing test. Have converted to draft while I fix it later today

mrchrisadams:
hi @tortila this is good for a quick review now. Danke!

templates/admin/accounts/datacenter/fieldset.html
```diff
@@ -54,3 +55,54 @@
...
+            <a href="https://www.thegreenwebfoundation.org/support-form/">
+            contact support.
+            </a>
+        </p>
+    {% endif %}
+
+    {% comment %} 
+        We have some providers listed in this DC, show them.
+    {% endcomment %}
+    {% if associated_providers_count < 1000 and associated_providers %}
```
tortila, line 76: I noticed that this condition already exists in the view, and is repeated here in the template. What is your main concern about: the performance implications or how the list is displayed? Is there's a way to address the concern in a single place in code?

tortila (commented):
Thanks for fixing this on such a short notice! The screenshots and description you provided are very helpful to understand the scope of changes. I'm happy to go ahead with merging this PR as soon as my questions about limiting the number of associated providers are answered

templates/admin/accounts/datacenter/fieldset.html
mrchrisadams: Good point. I've updated this accordingly.

apps/accounts/admin.py
```diff
@@ -915,6 +952,30 @@ def get_inlines(self, request, obj):
...
+            associated_providers_count = datacentre.hostingproviders.all().count()
+
+            if associated_providers_count > 1000:
+                extra_context["associated_providers_count"] = associated_providers_count
+                extra_context["associated_providers"] = []
+            else:
+                extra_context["associated_providers_count"] = associated_providers_count
+                extra_context[
+                    "associated_providers"
+                ] = datacentre.hostingproviders.all()
```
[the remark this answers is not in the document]
mrchrisadams: Good point. I was mainly thinking about the case where our admin really struggles to show profile pages that have loads of IP ranges.

Right now, while we have more than 1000 providers in our system, no datacentre is hosting more than a few hundred of them, so I've just removed the logic in the template and view code. 

Thanks Oliwia!

mrchrisadams:
hi @tortila - I've made the changes as you suggested.

mrchrisadams:
[Update listing of how providers list how they are associated with datacentres](https://trello.com/c/5vFOCJQ8/329-update-listing-of-how-providers-list-how-they-are-associated-with-datacentres)

PR #319, opened 2022-10-13
WIP - Update API to list more of the supply chain for a given resolved IP address. Fixes #316

### LINKED ISSUE #316: [ENHANCEMENT] Show the full range of evidence in our matches when looking up an domain or IP address
**Is your feature request related to a problem? Please describe.**

At present, when you look up a domain name or IP address with our greencheck service, we try to match the IP to a set of 'known green' IP ranges what we refer to as a _Green IP Range_.

If we can't find any matching IP ranges, we then do the same look up for the AS (the autonomous network) that the IP address is associated with - what we call a _Green ASN_.

These _Green IP Ranges_ and _Green ASNs_ are what we link to a specific organisation, and the uploaded supporting evidence of their sustainability claims, and where possible, we link to this so that anyone checking a website can easily see the basis we have used to mark it as green.

Presented visually it looks a bit like this:

1  How multiple IP matches work with our greencheck

This transparency is a key thing our API is designed to provide, to help inform people building digital services to help them choose greener infrastructure, but also to help end users and customers ask for the same.

The problem we have is that we only show the first match we see. So the results usually look like this:

2  How multiple IP matches work with our greencheck

When we have two sites that have the same IP, it's problematic, as it hides information from one provider.

3  How multiple IP matches work with our greencheck

**Describe the solution you'd like**
_A clear and concise description of what you want to happen._

There isn't a easy fix, but one thing we can do to make it possible to present information more correctly is to share the full set of _Green IP Range_ and _Green ASN_ matches in the API payload.

This would make it possible to update our own greencheck service at https://www.thegreenwebfoundation.org/green-web-check, but also allow any other users of our API to do the same.

Even if we don't change the the main greencheck, we can use this in our "detailed" check, to prototype how to present this information, for use in the more commonly used checking service on our main wordpress site.

More concretely, if an API response looked like this:

```json
{
  "url": "thegreenwebfoundation.org",
  "hosted_by": "Cloudflare",
  "hosted_by_website": "www.cloudflare.com",
  "partner": null,
  "green": true,
  "hosted_by_id": 779,
  "modified": "2022-10-12T11:21:57",
  "supporting_documents": [
    {
      "id": 18,
      "title": "Blog post - The Climate and Cloudflare",
      "link": "https://blog.cloudflare.com/the-climate-and-cloudflare/"
    },
    {
      "id": 21,
      "title": "Cloudflare 2020 Emissions Inventory",
      "link": "https://s3.nl-ams.scw.cloud/tgwf-web-app-live/uploads/Cloudflare_Emissions_Inventory_-_2020.pdf"
    }
  ]
}
```

We would have a payload that looked more like this:

```json

{
  "url": "thegreenwebfoundation.org",
  "hosted_by": "Cloudflare",
  "hosted_by_website": "www.cloudflare.com",
  "partner": null,
  "green": true,
  "hosted_by_id": 779,
  "modified": "2022-10-12T11:21:57",
  "supporting_documents": []
  "matching_providers": [
     provider_with_evidence_for_ip,
     provider_with_evidence_for_ip_range,
     provider_with_evidence_for_ASN,
]
}

```

This would increase the size of our payload, and but allow for showing more correct info in checks.

We might want to hide the extra info behind a GET param, or only make this available from the `api/v3/greencheck` endpoint.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered that might help serve as a possible solution.

1. With IPv6 we have probably have enough IP addresses for this not to be a problem, but support is not universal.
2. We are working on a domain based approach called carbon.txt, but this is not a thing we have officially launched yet.

**Additional context**

mrchrisadams:
This PR introduces an update to our v3 API, to expand the payload we return in API requests, to show the full list of matching providers instead of just the first one.

This paves the way to update the greencheck on our main website to do the same, which should help resolve issue #316, which provides more detail.

https://github.com/thegreenwebfoundation/admin-portal/issues/316

### Implementing this

To make this change, we will need to account for the way we perform lookups in a number of other places, as a change in the underlying greencheck logic will cascade through the following places:

1. looking up a the response for detail lookup we currently offer in the admin
2. how we build an API response for the quick lookups from the green domain 'cache' table
3. how we build an API response for the extended lookups when we skip the green domain cache table
4. how we return results for the workers in the message queue writing to the greencheck log table
5. how we build the image badges

### DIFF
--- a/apps/greencheck/api/legacy_views.py
+++ b/apps/greencheck/api/legacy_views.py
@@ -16,7 +16,17 @@ from ..serializers import GreenDomainSerializer
 
 logger = logging.getLogger(__name__)
 
+from ..viewsets import AbstractGreenDomainViewset
+
 checker = GreenDomainChecker()
+
+
+class GreenDomainViewset(AbstractGreenDomainViewset):
+    """
+    The legacy version of our API, served at the older `greencheck` routes, with
+    """
+
+    pass
 
 
 def augmented_greencheck(check):
--- a/apps/greencheck/domain_check.py
+++ b/apps/greencheck/domain_check.py
@@ -26,6 +26,8 @@ from ipwhois.net import Net
 from .choices import GreenlistChoice
 from .models import GreenDomain, SiteCheck
 
+from typing import List
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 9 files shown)


mrchrisadams:
Closing this PR, as the initial issue has been closed, and we are addressing it with the carbon.txt work instead.

PR #322, opened 2022-10-26
ProviderRequest: basic views & admin

### LINKED ISSUE #311: [FEATURE] Authenticated users are able to access their submissions
**User story:**
Given a feature toggle is turned on for my particular account,
as an authenticated user
when I access the submissions page,
then I'm able to see all submissions related to my account.

**Acceptance criteria:**
- `/submissions` page (name to be confirmed) is accessible by selected users
- all submissions for the current user are shown on that page
- a feature flag enabling/disabling this access is managed by staff in the admin panel
- each submission has some data related to it (to be defined) and a status
- submissions panel is accessible by staff in the admin panel, sorted by `pending` submissions first 

**Developer notes:**
- This is the first task related to the new supplier portal. The focus is to model the data on the backend and make minimal changes on the front-end to expose this functionality to selected users (staff for now). 
- As pointed out by Chris, we can utilize [`django-waffle`](https://waffle.readthedocs.io/en/stable/index.html) to configure this new functionality behind a feature flag.
- We need to decide on which domain to expose this new functionality, for example: `portal.greenwebfoundation.org`, `app.greenwebfoundation.org`, etc.

tortila:
Closes #311 

Scope of changes:
- new models for `ProviderRequest` and related sub-resources, that model the data input from the providers to our system,
- admin panel for managing new and existing provider requests, accessible by staff,
- basic list and detail view with access managed by `django-waffle` flag,
- explicit permissions for the "admin" group to manage provider requests in the admin panel,
- tests for custom validation of the green evidence attached to the provider request, and who should have access to admin panel

Some corners have been cut:

- no new domain configured: the new view is served on the same domain as admin, under the `/requests/` path


apps/accounts/models/provider_request.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ProviderRequestIPRange(models.Model):
+    start = IpAddressField()
+    end = IpAddressField()
+    location = models.ForeignKey(ProviderRequestLocation, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f"{self.start} - {self.end}"
+
+    def clean(self):
```
mrchrisadams, line 84: If we don't already have tests for the IP validation working, it would be nice to see them here. If we do, it's less of a priority.

I think our IPAddressField is a one we built ourselves rather than being from the django framework. If we don't have tests there, then I think it may be worth adding them at that level.

apps/accounts/models/provider_request.py
```diff
@@ -0,0 +1,111 @@
...
+        validate_ip_range(self.start, self.end)
+
+
+class EvidenceType(models.TextChoices):
```
mrchrisadams, line 88: A docstring about the expected usage of this evidence would be helpful. Who is uploading the evidence, and where are we intending to surface it to end users?

apps/accounts/models/provider_request.py
```diff
@@ -0,0 +1,111 @@
...
+    title = models.CharField(max_length=255)
+    link = models.URLField(null=True, blank=True)
+    file = models.FileField(null=True, blank=True)
+    location = models.ForeignKey(ProviderRequestLocation, on_delete=models.CASCADE)
+    type = models.CharField(choices=EvidenceType.choices, max_length=255)
+
+    def __str__(self):
+        return f"{self.title} ({self.type})"
+
+    def clean(self):
```
mrchrisadams, line 104: Could we have some tests on this logic please? It's 'our' code, and I don't think we have tests on the earlier implementation of this logic.

apps/accounts/models/provider_request.py
```diff
@@ -0,0 +1,111 @@
...
+        validate_ip_range(self.start, self.end)
+
+
+class EvidenceType(models.TextChoices):
+    ANNUAL_REPORT = "Annual report"
+    WEB_PAGE = "Web page"
+    CERTIFICATE = "Certificate"
+
+
+class ProviderRequestEvidence(models.Model):
```
mrchrisadams, line 94: As discussed in person about the docstrings. Having one or two lines about the intended use would be helpful for our future selves.

apps/greencheck/validators.py
```diff
@@ -0,0 +1,16 @@
+import ipaddress
+from django.core import exceptions
+
+
+def validate_ip_range(ip_start, ip_end):
```
mrchrisadams, line 5: YAY! No needless use of classes! 👍

mrchrisadams (approved):
Hi Oliwia, I'm happy with the logic on this - no changes are required there.

The only thing would be testing and docstrings as we discussed in person. 

You're free to merge this in after adding them, the only thing I think of here is whether we need to hold off on showing the ProviderRequest to external users who might be logging into our admin.

apps/accounts/models/provider_request.py
tortila: There's already a test for the IP validation in another place - the hostingprovider admin (which uses the same code)

mrchrisadams:
hi Oliwia,  sorry about this - I'm happy to merge this in, but there's a pipfile.lock conflict. Can you resolve that? 

Once it's done, I'll be able to merge into master.

PR #324, opened 2022-10-28
Add sentry tracing for investigating slow requests

mrchrisadams:
Hi @tortila I had a quick look at adding "spans and traces" support into this app, and it turns out to be a 1-2 line change.

This PR introduces the change, and documentation in the config file, but until an environment variable is set, there will be no changes to existing functionality, and no trace information will be sent.

I figured adding this would mean that we can:

1. catch behaviour in staging as we develop the new work
2. check out any per issues with prod traffic,  assuming we sample sensibly, or even split how traffic going to admin.TGWF differently is dealt with versus api.TGWF (api receives literally 1000x the traffic).

C


tortila (approved):
Good idea, thanks for enabling this!

PR #330, opened 2022-11-01
Link to dev docs site from API docs

fershad:
This PR adds a simple banner to the top of the `/api-docs/` page which links out to the new developer documentation website.

This gives exposure to the new documentation site, which provides more user-friendly API guidance. Although the new site currently only covers two APIs (Greencheck and IP to CO2), it is expected that will expand to cover all the major Green Web Foundation APIs.

### DIFF
--- a/templates/partials/docs_banner.html
+++ b/templates/partials/docs_banner.html
@@ -7,8 +7,8 @@
 
 <style>
     div.announce-banner {
-        background-color: #466D22;
-        color: #FFF;
+        background-color: #EAF5E0;
+        border: #65B54C 1px solid;
         width: max-content;
         margin: 1.5rem auto;
         padding: 1rem 1.5rem 1.5rem;
--- a/templates/partials/docs_banner.html
+++ b/templates/partials/docs_banner.html
@@ -1,0 +1,39 @@
+{% load static %}
+
+<div class="announce-banner">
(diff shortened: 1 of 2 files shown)


mrchrisadams:
Hey fish this works totally works, danke !

I have one tiny request tho - would it be possible to done down the green a tad and use a lighter version like the kind we have defined in our Figma colors doc?

At present it looks like this:

Screenshot 2022-11-03 at 07 09 09

I think it's a bit less jarring if we us a similar text color, but use one of the lighter tints instead, like so:

Screenshot 2022-11-03 at 07 19 55

For this example, I changes the background shade to the green 800 tint in the figma file, and used the same green as is visible in the logo green -500 

```css

background-color: #EAF5E0; # green 800
border: #65B54C 1px solid; # green 500
```

For reference the figma file is here:
https://www.figma.com/file/VdYdk14woWHD2ov9nfcHw6/TGWF-Colors?node-id=0%3A1

If you're okay to make that change, I'm very happy for you to merge this in.

If it turns out to be a faff, I'm happy to merge this in anyway, as I expect we'll end up working with a new figma reference file now that we have the new brand almost finished

fershad:
Cheers @mrchrisadams I've made that change. Can't merge PRs into this repo (probably for the better) so I'll leave it with you.

PR #333, opened 2022-11-03
First pass at directory filtering

mrchrisadams:
This is a work in progress demo of setting up our filtering.

If you visit `/directory`, you'll see a filterable view, allowing you to filter by country, service provided and provider name.

This work is behind a the `directory_listing` feature flag.

### Where I'd appreciate some help

I've done almost no work to try to speed up queries, and as as result it's not fast.

I think this is because for every provider on a page, we do more queries to fetch the tags and the names, when we could get them just one with `prefetch_related`. It's not a thing I'm so familiar  with, so I'd appreciate someone pairing with me to figure it out.

### DIFF
--- a/.github/workflows/run-deploy-staging.yml
+++ b/.github/workflows/run-deploy-staging.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Run playbook
         uses: dawidd6/action-ansible-playbook@v2
         with:
-          playbook: ./ansible/deploy-staging.yml
+          playbook: ./ansible/deploy.yml
           directory: ./
           key: ${{secrets.SSH_PRIVATE_KEY}}
           inventory: |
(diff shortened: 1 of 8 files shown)


mrchrisadams:
Ah, slight update. it looks like django taggit doesn't support `prefetch_related` if you use the  `names()` or `slugs() method like i did in the template. Maybe this in this case we should use `all()`

More below:

https://`github.com/jazzband/django-taggit/issues/687

mrchrisadams:
this PR also has no flags added. if we push this to master it would make sense to add them, obvs

tortila (approved):
> this PR also has no flags added

Do you mean to hide the new view behind a feature flag? I think it's a good idea. Otherwise :+1:, good to see that this proof of concept works

PR #349, opened 2022-12-07
Support multiple active filters. Fixes #348

### LINKED ISSUE #348: [ENHANCEMENT]  Allow filtering by multiple choices at a time in django admin
Right now, if you're a member of staff, you can filter the admin view based on a single filter at a time, as is normal with django.

However, you often want to filter by more than one at the same time - for example to return all hosting providers  marked as 'needs-follow-up' AND 'no-user'.

The library below supports these kinds of queries:

https://github.com/ctxis/django-admin-multiple-choice-list-filter

mrchrisadams:
This PR adds support for using multiple filters when filtering hosting providers up to a maximum of 4 simultaneous tags.

Previously our filter view looked like this:

Now we have an option to progressive filter by more tags, up to a maximum of four tags at the same time. 

This is an AND query, not an OR query, so it will only match items that have all the active tags. 

When more than four filters are in active use a warning message is shown and advises users how to reactivate the filtering.

This code is not beautiful, but it seems to work.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -87,6 +87,7 @@ duckdb = "==0.5.0"
 dateutils = "*"
 django-filter = "*"
 django-sql-explorer = "*"
+django-admin-multiple-choice-list-filter = "*"
 
 [pipenv]
 # Needed for `black`. See https://github.com/microsoft/vscode-python/pull/5967.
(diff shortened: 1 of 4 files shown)


mrchrisadams:
@tortila would you please give this a quick look over and merge in if you don't see any real clangers?

You can see it in action on the staging site and play around with it if need be.

PR #360, opened 2022-12-19
New registration button links to form start

hanopcan:
Please have a look at my new link to make the start of the reg process actually a thing


mrchrisadams:
Thanks Han!

PR #362, opened 2022-12-19
Add service and required tagged item base model

mrchrisadams:
Hey Oliwia / @tortila 

One thing that came up when speaking to Hannah today was the set of services that providers list themselves as offering.

We currently use the basic Taggit 'Tag' model, but given that each service has a description, I'm wondering if we might be better off making the move sooner rather than later, as we maintain a list of services anyway at the link below.

https://www.thegreenwebfoundation.org/directory/services-offered/

I can imagine scenarios where we have specific urls like `/service/object-storage` in future as a way to help people understand what specific services are, and who different providers might be.

Also, given that all our forms would likely need these listed in tests, and in development, if we're going to add a migration to list them, it might be worth adding the new models too. We already use them for the internal staff labels, and making these new models was really quick.

What do you think? I'm parking this here to discuss in future.

### DIFF
--- a/apps/accounts/models/hosting.py
+++ b/apps/accounts/models/hosting.py
@@ -156,9 +156,42 @@ class ProviderLabel(tag_models.TaggedItemBase):
     to help us categorise and segment providers.
     """
 
-    content_object = models.ForeignKey("Hostingprovider", on_delete=models.CASCADE,)
+    content_object = models.ForeignKey(
+        "Hostingprovider",
+        on_delete=models.CASCADE,
+    )
     tag = models.ForeignKey(
-        Label, related_name="%(app_label)s_%(class)s_items", on_delete=models.CASCADE,
+        Label,
+        related_name="%(app_label)s_%(class)s_items",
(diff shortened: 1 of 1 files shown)


tortila:
Hey Chris, I have some follow-up questions. Is this PR considered work in progress? I'm having some troubles correlating the change set and what you wrote in the description. I'd appreciate if you could help me clarify: what is expected outcome of this PR? 

And could you help me understand what you mean here?
> Also, given that all our forms would likely need these listed in tests, and in development, if we're going to add a migration to list them, it might be worth adding the new models too. We already use them for the internal staff labels, and making these new models was really quick.

We already have tests that rely on a list of services being available, e.g. https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/accounts/tests/test_provider_request.py#L49. 
In general, for development we don't use any data fixtures at all so far, right? And what type of migration do you have in mind?

mrchrisadams:
Closing this - we delivered it in another #467

PR #367, opened 2022-12-30
Hs request form styling

hanopcan:
Styling for request form. I think we can merge this straight in as it's not live and makes it easier for everyone to see and test.


hanopcan:
Hi @mrchrisadams Can you take a look at this pull request for me please? The thing blocking me from merging this in is that it is failing the checks. I'm not sure why and I suspect it might be something you'll look at and diagnose relatively quickly. Thanks!

tortila:
@hanopcan the issue should be fixed now - the check has passed.

What was wrong: `pipenv` (the tool we use for managing dependencies) complained that there's a mismatch for a hash of one of the dependencies. I'm not sure why, since you didn't add any new packages in this PR. My only explanation is that something had to change in the upstream.

What I did: I ran the command `pipenv lock` manually, this updated the file `Pipfile.lock` with new hashes as seen in the PyPI registry. I pushed the changes in this commit: https://github.com/thegreenwebfoundation/admin-portal/pull/367/commits/8c914e36cd1119615ff1f24f3e716051a3b9c32d. `pipenv` no longer complains!

tortila (approved):
Nicely done @hanopcan! It looks great. 

From my perspective it's OK to merge this in, any further work can be added later (e.g. fixing the IP ranges validation)

PR #372, opened 2023-01-18
Staging basic auth

mrchrisadams:
This PR adds Basic auth checking for use in staging environments.

It introduces three new environment variables with what should be helpful defaults that we can override if need be:

```
BASICAUTH_DISABLE=(bool, True),
BASICAUTH_USER=(str, "staging_user"),
BASICAUTH_PASSWORD=(str, "strong_password"),
```

We also have a few basic tests to exercise this code. 

Why have this?

From painful experience when debugging issues, it's really helpful to sanity check why changes might not be reflecting on a site, and if a link to a staging site is shared by accident to external users, it saves on support questions.

There is an argument about being able to control who sees what we're doing but given we're an open source codebase, that's much less of an issue.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -91,6 +91,7 @@ django-admin-multiple-choice-list-filter = "*"
 django-formtools = "*"
 django-betterforms = "*"
 django-convenient-formsets = "*"
+django-basicauth = "*"
 
 [pipenv]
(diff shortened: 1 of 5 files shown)


tortila (approved):
Looks good!

Issue #336, opened 2022-11-16
Introduce "locations" that are assigned to hosting providers


tortila:
In the future we want to enable multiple locations per provider. 
Location should include: 
- country,
- city, 
- name, 
- possibly a geocoded point (to allow displaying on the map)

This task includes:
- modifying Hostingprovider model (remove country field), create a new Location model (that contains the country field) with a reference to the Hostingprovider
- migration of existing data

tortila:
![](https://github.trello.services/images/mini-trello-icon.png) [Introduce "locations" that are assigned to hosting providers](https://trello.com/c/gFRRi3B1/43-introduce-locations-that-are-assigned-to-hosting-providers)

mrchrisadams:
hey @tortila can I check what we'd need here?

We'd need to extend the `ProviderRegistrationView` with a something like a new entry in the FORMS list.

So here,

```py
FORMS = [
        (Steps.ORG_DETAILS.value, OrgDetailsForm),
        (Steps.SERVICES.value, ServicesForm),
        (Steps.GREEN_EVIDENCE.value, GreenEvidenceForm),
        (Steps.NETWORK_FOOTPRINT.value, NetworkFootprintForm),
        (Steps.CONSENT.value, ConsentForm),
    ]
```

We'd have something like this:

```py
FORMS = [
        (Steps.ORG_DETAILS.value, OrgDetailsForm),
        (Steps.LOCATIONS.value, OrgLocationForm),
        (Steps.SERVICES.value, ServicesForm),
        (Steps.GREEN_EVIDENCE.value, GreenEvidenceForm),
        (Steps.NETWORK_FOOTPRINT.value, NetworkFootprintForm),
        (Steps.CONSENT.value, ConsentForm),
    ]
```

And update the templates accordingly:

```py
TEMPLATES = {
        Steps.ORG_DETAILS.value: "provider_registration/about_org.html",
        Steps.LOCATIONS.value: "provider_registration/locations.html",
        Steps.SERVICES.value: "provider_registration/services.html",
        Steps.GREEN_EVIDENCE.value: "provider_registration/evidence.html",
        Steps.NETWORK_FOOTPRINT.value: "provider_registration/network_footprint.html",
        Steps.CONSENT.value: "provider_registration/consent.html",
    }
```

And then the `done ` method to deal with the form processing. This about right?

tortila:
[...]

PR #378, opened 2023-01-18
Adding a new screen for multiple locations

mrchrisadams:
This is a PR for adding a new step in our form asking users to add one or more location that they operate in.

It was done to get an idea of how hard it was to add new steps to the form as much as anything else, and act as a starting point for discussions about how we ask about locations.

### TODO

- [x] revisit language in the form
- [x] fix the tests
- [x] figure out how this should look in the preview / summary
- [x] ~~Decide about django cities~~

This leaves the preview out for now, as its being implemented in another ticket.

### Post approval todo

- [ ] run the necessary migrations after merging in

#### Key changes made

**Additions**

We now have a locations step in our Verification Request form wizard. This is reflected in the locations template, a new `LocationForm`, and an updated location model, where we list the optional name and description.

**Changes**

**hosting provider model:**  - we now have a city field for easy import from datacentres, and including locations
**request location** - previously we represented the location in the provide request as fields in our form. These are extracted into a separate object in the form, that point to the model we defined before.
**verification request form** - the form has new logic to account for there being one or more locations in in the request, and this shows in the summary (I haven't done much here though, it's very minimal)

**Deletions**

**We no longer have the city and country on the first step of the form.**

### Things to think about

**Do we use django cities?**

There is a full list of cities in a django package here that has information about their rough location, host country and the rest. This would make geolocating and putting stuff on maps muuuch nicer, as long as we can clearly communicate that we only need approximate location.

https://pypi.org/project/django-cities/

_Django cities looks like it would allow us to just ask for one field, which would then give us country, region, etc., but implementing it would be a bit bigger than this ticket would allow. One for the next version._


mrchrisadams:
Hey @tortila, I think this PR  is good for review now. There are some migrations needed

apps/accounts/models/provider_request.py
```diff
@@ -99,24 +102,14 @@ class ProviderRequestLocation(models.Model):
     in which the new provider offers services.
     """
 
+    name = models.CharField(max_length=255, blank=True)
+    description = models.CharField(max_length=255, blank=True)
```
tortila, line 106: Why is the new `description` field added to the location object? It's surprising to me because we have agreed to "name", "city" and "country" before (see https://github.com/thegreenwebfoundation/admin-portal/issues/336). What is the description of a location useful for?

apps/accounts/templates/provider_registration/locations.html
```diff
@@ -0,0 +1,114 @@
...
+    <div class="convenient-form bg-neutral-200 rounded-xl my-6 p-4">
+        <div class="form__container--location">
+            {{ form.non_field_errors }}
+
+            {% comment %}
+            Iteration over all fields in the location form. This is
+            equivalent to listing a span for each form.field_name
+            where field_name is in: ["title", "link", "file", "type",
+            "public", "description"]
+            {% endcomment %}
```
tortila, line 62: Small nitpick: these fields `[["title", "link", "file", "type", "public", "description"]`are for evidence and not for location. This comment can be either removed or updated

apps/accounts/forms.py
```diff
@@ -384,16 +391,84 @@ class ConsentForm(forms.ModelForm):
...
+
+    country = CountryField().formfield(
+        label="Country",
+        blank_label="Select country",
+        help_text="Choose a country from the list.",
+    )
+
+    class Meta:
+        model = ac_models.ProviderRequestLocation
+        # exclude = ["request"]
```
tortila, line 460: It's safe to remove this commented-out line (`fields` and `exclude` are mutually exlusive)

tortila (commented):
Thanks for working on this @mrchrisadams! I reviewed the code now and also tested it manually - looks good! 
I left some comments in which I'm requesting a bit of cleaning up. I'm also wondering about the new field in the location object. Could you please respond to the comments before this is merged?

mrchrisadams:
Hi @tortila I've made the requested changes. Would you give the go ahead if you're happy? I'll take care of merging and running migrations as agreed.

tortila (approved):
LGTM :+1: 
Nice work with the partial template, looks much cleaner now.

Feel free to proceed with merging and running migrations after the deployment is done

PR #380, opened 2023-01-20
Validate IP addresses in the registration form correctly

tortila:
Scope of changes:
- changed the type of the field for IP addresses in the form to [`GenericIPAddressField`](https://docs.djangoproject.com/en/4.1/ref/forms/fields/#genericipaddressfield) for both: IP start and end. This is a built in type that provides the validation for IPv4 and IPv6 addresses,
- added a small change in the `ProviderRequestIPRange.clean` so that the validation provided by `GenericIPAddressField` kicks in properly + documented the unexpected behavior,
- added form tests that cover edge cases,
- minor changes in the wording on the IP ranges form ("delete" button is tied to a single IP range, not multiple)

### DIFF
--- a/apps/accounts/forms.py
+++ b/apps/accounts/forms.py
@@ -283,7 +283,7 @@ class CredentialForm(forms.ModelForm):
         fields = ["type", "title", "link", "file", "description", "public"]
         help_texts = {
             "type": "What kind of evidence are you adding? Choose from the dropdown list.",
-			"title": "Give this piece of evidence a title.",
+            "title": "Give this piece of evidence a title.",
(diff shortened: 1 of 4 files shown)


tortila:
[On entering an invalid IP range, we need to handle errors better](https://trello.com/c/09Ft16kN/72-on-entering-an-invalid-ip-range-we-need-to-handle-errors-better)

mrchrisadams (approved):
Hi @tortila thanks, I'm happy with this.

It wasn't obvous in the issue what was causing this, but the linked trello card provided the info, and I've been able to test the behaviour locally as well. Please merge this in when you're ready.

PR #386, opened 2023-01-30
Hs login form

hanopcan:


hanopcan:
Hi @mrchrisadams

I've done some work on styling the registration and login forms today. I haven't merged this request into master yet as I think that `/registration` is currently the live registration form. I'm not 100% sure this is being used by users, but I think it might be.

If it is being used, I guess there are a few things to check before this unwittingly goes live:

1. Have I broken anything and is signup still working as expected?
2. Should I get rid of the login link at the bottom of the page because it directs users to the new login page which isn't ready to go live yet.
3. Can we live with the registration form styling being completely different to everything else a user will see (I think we can live with this, it's not ideal, but the sign up page will be used so infrequently I think a user will forget about it quite quickly once they are in.

Any other thoughts are welcome.

hanopcan:
https://trello.com/c/HDIVh12D/91-first-pass-at-styling-login-and-sign-up-forms

mrchrisadams:
Hi han, thanks for this - it looks good!

Did you want me to try resolving the merge conflict in the CSS, or woudl you prefer I had a go?

> - Have I broken anything and is signup still working as expected?

From what I can tell all seems to be ok!

> - Should I get rid of the login link at the bottom of the page because it directs users to the new login page which isn't ready to go live yet.

I think it's ok to have this TBH. It's still the same name, and if we need to, creating a persistent message to explain that this really is the same site is a thing we can do if we see people getting too confused.

> - Can we live with the registration form styling being completely different to everything else a user will see (I think we can live with this, it's not ideal, but the sign up page will be used so infrequently I think a user will forget about it quite quickly once they are in.

Yes, I think this is ok - the alternative is designing lots of logic to hide the new design and I don't think it justifies the investment of time to do so.

mrchrisadams (approved):

hanopcan:
@mrchrisadams 

Leave the merge conflicts with me to resolve. I'm happy to sort that, will be quick for me to do.

Thank you for the feedback on the other bits! 

> I think it's ok to have this TBH. It's still the same name, and if we need to, creating a persistent message to explain that this really is the same site is a thing we can do if we see people getting too confused.

I think there probably is another action to put some messaging in place ahead of time, and I think maybe in the new portal we should include a link to the old portal. We kinda touched on that today in our discussion with Justine. I think that's a separate task to add into scope. Won't be hard and it's basically text changes. Not a blocking change.

PR #389, opened 2023-02-01
Fix mac dependencies in development

mrchrisadams:
On a clean mac OS X install, duckdb 0.5.0 was causing a failing build.

This fixes it.


mrchrisadams:
OK, this failure is related to a new version of pytest using exceptiongroups which python 3.8.10 doesn't support without a polyfill.

PR #402, opened 2023-02-08
Send email notification to support staff and user when submitting a form

mrchrisadams:
This PR introduces the logic to send an email to:

1. our internal support address
2. the user who submitted the form

The email message contains a link to the submission for reference along with a helpful message.

- [x] email uses a text template we can read and change the copy
- [x] email uses an html template we can read and change the copy
- [x] email links back to to the provider, and includes relevant details
- [x] an email is sent to the submission author and the support staff
- [x] we have fail gracefully when email is not working, with a intelligible error message

### DIFF
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -279,7 +279,7 @@ class ProviderRegistrationView(LoginRequiredMixin, WaffleFlagMixin, SessionWizar
         consent.request = pr
         consent.save()
 
-
+        # send an email notification to the author and green web staff
         self._send_notification_email(self.request.user, pr)
 
         return redirect(pr)
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -1,6 +1,8 @@
 import waffle
 from waffle.mixins import WaffleFlagMixin
 from enum import Enum
+
+import smtplib
 from django.core.files.storage import DefaultStorage
 from django.contrib import messages
 from django.contrib.auth.models import Group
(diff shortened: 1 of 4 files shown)


mrchrisadams:
[Send an email when a new verification request is submitted](https://trello.com/c/vOtn6U2m/78-send-an-email-when-a-new-verification-request-is-submitted)

apps/accounts/views.py
```diff
@@ -271,6 +279,9 @@ def done(self, form_list, form_dict, **kwargs):
         consent.request = pr
         consent.save()
 
+        # send an email notification to the author and green web staff
+        self._send_notification_email(self.request.user, pr)
```
tortila, line 283: Since `_send_notification_email` is defined in the same class, there's no need to pass `self.request.user` as a parameter - the user object can be retrieved inside `_send_notification_email`, since it has access to `self`.

apps/accounts/views.py
```diff
@@ -315,3 +326,40 @@ def get_context_data(self, form, **kwargs):
...
+    def _send_notification_email(self, user: User, provider_request: ProviderRequest):
+        """
+        Send notification to support staff, and the user to acknowledge their submission
+        """
+
+        current_site = get_current_site(self.request)
+
+        request_path = reverse("provider_request_detail", args=[provider_request.id])
+
+        link_to_verification_request = f"{self.request.scheme}://{current_site.domain}/{request_path}"
```
tortila, line 339: `reverse` will return an absolute path, so `request_path` already starts with a `/`. Then I think that the URL needs to look like this: `f"{self.request.scheme}://{current_site.domain}{request_path}"`. Could you add an assertion for the URL included in the email in the test that you wrote to verify this please?

mrchrisadams:
hey @tortila it looks like you're reviewing this PR - can you please tag me when you're done? I'll make the requested changes once you have.

PR #405, opened 2023-02-14
Add extra option to allow listing why you are not listing your own IP Ranges or AS Numbers to allocate

mrchrisadams:
We have a few scenarios where a provider might not list the IP ranges or AS numbers in a verification request.

Some providers don't explicitly provide exclusive use of IP ranges for example, and for others they might change so often that it makes more sense to send updates via an API instead.

This PR introduces a new option on the network footprint step of the form allowing a user to list this information.


mrchrisadams:
[Verification request form - changes to network footprint page - add checkbox options](https://trello.com/c/bAZV4UQl/83-verification-request-form-changes-to-network-footprint-page-add-checkbox-options)

apps/accounts/forms.py
```diff
@@ -21,6 +21,10 @@
 from .utils import tags_choices
 from django.utils.safestring import mark_safe
 
+
+import logging
+logger = logging.getLogger(__name__)
+
```
tortila, line 27: logging is not used in this module - this can be removed

apps/accounts/forms.py
```diff
@@ -377,8 +393,38 @@ class NetworkFootprintForm(MultiModelForm):
...
+        super().clean()
+
+        # fetch our forms
+        ip_form, asn_form, explanation_form = [form for form in self.forms.values()]
+
+        # is there any description we can read?
+        explanation_present = any(explanation_form.cleaned_data.values())
+
+        # do we have any in the two forms asking for network data?
+        network_data_present = ip_form.forms and asn_form.forms
```
tortila, line 415: I think that an `or` condition is more appropriate here: we expect that either some IP ranges or ASNs are present. I also suggest we add a test case for that (IP form present, no ASN form present - or vice versa)

apps/accounts/forms.py
```diff
@@ -377,8 +393,38 @@ class NetworkFootprintForm(MultiModelForm):
...
+        """
+        Validate that a form at least has:
+        1. some IP and/or AS data
+        2. or some explanation
+
+        """
+        super().clean()
+
+        # fetch our forms
+        ip_form, asn_form, explanation_form = [form for form in self.forms.values()]
```
tortila, line 409: I'm not sure if `self.forms` respects the ordering, maybe we shouldn't rely on it. It would be better to get each form by the key

apps/accounts/tests/test_form.py
```diff
@@ -167,3 +168,112 @@ def test_ip_range_form_validation(form_data):
...
+        "ips-INITIAL_FORMS": "0",
+        "asns-TOTAL_FORMS": "0",
+        "asns-INITIAL_FORMS": "0",
+        "extra-description": ""
+    }
+
+
+class TestNetworkFootprintForm:
+
+    @pytest.mark.skip(reason="Not now")
```
tortila, line 236: Is there any reason why this test is skipped?

apps/accounts/tests/test_form.py
```diff
@@ -167,3 +168,112 @@ def test_ip_range_form_validation(form_data):
...
+        # given a valid submission
+        multiform = NetworkFootprintForm(wizard_form_network_data)
+
+        # when: a validation check has been run
+
+        validation_result = multiform.is_valid()
+
+        assert validation_result is True
+
+    @pytest.mark.skip(reason="Not now")
```
tortila, line 248: Same here - does this test need to be skipped?

apps/accounts/forms.py
```diff
@@ -357,6 +354,17 @@ class Meta:
 AsnFormset = forms.formset_factory(AsnForm, formset=MoreConvenientFormset, extra=0)
 
 
+class ExtraNetworkForm(forms.Form):
```
tortila, line 357: It just occurred to me that at this point we're accepting the text input in the newly added form (explanation), but do not persist it. Which means it will not be accessible by anyone after the form is submitted.

If you want to make the explanation available to GWF staff, you should add some logic in the view to persist this data

mrchrisadams: Duh! I'll add that extra line and migration file now

mrchrisadams:
hi @tortila this is the first PR to merge in - it's mostly stuff you have reviewed once already.

I'll work on tidying up the others now as well

apps/accounts/tests/test_provider_request.py
```diff
@@ -575,3 +594,42 @@ def test_approve_creates_evidence_documents(db):
...
+        wizard_form_preview,
+    ]
+    # when: a multi step submission has been successfully completed
+    response = _create_provider_request(client, form_data)
+
+    # then: a ProviderRequest object exists in the db
+    pr = response.context_data["providerrequest"]
+    assert models.ProviderRequest.objects.filter(id=pr.id).exists()
+
+    # then we have our explanation saved to the provider
```
tortila, line 631: Small suggestion: if you want to check that the explanation is saved to the provider, then it makes sense to compare the expected value to the saved object (`models.ProviderRequest.objects.filter(id=pr.id).get()`) rather than the body of the HTTP response.

tortila (approved):
I left one comment with regards to testing. The implementation of the new form and the validation logic looks good :+1:

mrchrisadams:
Thanks @tortila  I'll merge in once the tests show as green for this one.

PR #409, opened 2023-02-17
Add checks for empty formsets

mrchrisadams:
This PR introduces ~~a new class that we use~~ _validation features  of the django formset API_ in the wizard to ensure that in some steps where we support multiple submissions in a formset or multiform, we make sure we have at least one form submission before we proceed through the wizard.

This covers cases where our validation was already checking that location or submitted pieces of evidence were valid, but we were not checking that at least one had been submitted at that stage.

### DIFF
--- a/apps/accounts/forms.py
+++ b/apps/accounts/forms.py
@@ -328,25 +328,6 @@ class MoreConvenientFormset(ConvenientBaseFormSet):
         return seen
 
 
-class AtLeastOneSubmissionFormset(MoreConvenientFormset):
-    """
-    "A variant of our MoreConvenientFormset that validates that there is
-    at *least one* entry in a formset. This catches submissions in our
-    wizard form where we need to see at least one location, piece of evidence
-    and so on.
-    """
-
-    def clean(self):
-        """
-        Act like MoreConvenientFormset.clean, but if we haven't seen any
-        form submissions in our formset either, then we raise an validation
-        error
-        """
-        seen = super().clean()
-
-        if not seen:
-            raise ValidationError("There needs to be at least one submission", code="no_submissions_in_formset")
-
 
 # Part of multi-step registration form (screen 3).
 # Uses ConvenientBaseFormSet to display add/delete buttons
@@ -354,7 +335,9 @@ class AtLeastOneSubmissionFormset(MoreConvenientFormset):
 GreenEvidenceForm = forms.formset_factory(
     CredentialForm,
     extra=1,
-    formset=AtLeastOneSubmissionFormset,
+    formset=MoreConvenientFormset,
+    validate_min=True,
+    min_num=1,
 )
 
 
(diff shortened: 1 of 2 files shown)


apps/accounts/tests/test_form.py
```diff
@@ -274,3 +282,207 @@ def test_form_not_valid_if_no_explanation_or_network_data(
...
+        file_data = {
+            "file": SimpleUploadedFile(fake_evidence.name, fake_evidence.read())
+        }
+        form = CredentialForm(valid_uploaded_doc_data, file_data)
+        valid_result = form.is_valid()
+
+        assert not valid_result
+
+
+class TestGreenEvidenceForm:
```
tortila, line 349: I see that you created new classes in order to group tests related to a specific form together. This confused me initially, because in other places in the codebase we use classes to designate Django-style (and unittest-style) [`TestCase`](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.TestCase)-derived tests. If this additional level of organization is not strictly necessary, I suggest we stick to a pytest-style flat structure of test functions.

apps/accounts/tests/test_form.py
```diff
@@ -274,3 +282,207 @@ def test_form_not_valid_if_no_explanation_or_network_data(
...
+        }
+        file_data = {
+            "file": SimpleUploadedFile(fake_evidence.name, fake_evidence.read())
+        }
+
+        form = CredentialForm(valid_uploaded_doc_data, file_data)
+        valid_result = form.is_valid()
+        assert valid_result
+
+    def test_valid_with_file_credential(self, fake_evidence):
```
tortila, line 332: This one duplicates the name of the test case above

apps/accounts/tests/test_form.py
```diff
@@ -274,3 +282,207 @@ def test_form_not_valid_if_no_explanation_or_network_data(
...
+
+
+class TestLocationsForm:
+    """
+    Test that our evidence form step checks for at least one piece
+    of valid evidence exists - not just that the pieces of evidence
+    are themselves valid.
+    """
+
+    def test_valid_with_evidence(self, wizard_form_org_location_data):
```
tortila, line 466: wrong name of the test case?

apps/accounts/forms.py
```diff
@@ -322,6 +325,9 @@ def clean(self):
                 form.add_error(None, e)
             seen.append(form.cleaned_data)
 
+        return seen
```
tortila, line 328: This method should not return anything

apps/accounts/tests/test_form.py
mrchrisadams: I've made this change for this PR, but can we discuss this on Wednesday?

We have something like 66 classes containing tests that use pytest directly, and none of them inherit directly from any classes like `TestCase` - they're simply containers for organisation.

For what it's worth I've been loosely following this approach outlined in these conventions:

https://github.com/octoenergy/conventions/blob/master/python.md#test-folder-structure

However, I'm v happy to get an explicit convention documented that we can refer to for future PRs, becuase we have a mixture of both in the codebase, and no written guidance in our own docs right now.

mrchrisadams:
Closing this in favour of https://docs.google.com/spreadsheets/d/1nRSKHoTl61SjSsl5hYCgTcZQ79kyXLLrwCNiybFc1Po/edit#gid=0

PR #413, opened 2023-02-22
Fix file upload from the verification request form

tortila:
Small set of important bug fixes:
- uploading of the files through the verification request form was broken because `upload_to` argument was missing,
- accessing of the files uploaded through the verification request form was broken on local machines, because the `MEDIA_URL` was not served by Django,
- uploading files with the same name (even by different users) was overwriting the old files, potentially leading to exposing confidential documents - this is fixed by setting `AWS_S3_FILE_OVERWRITE` flag to `False`.

Tested locally (local filesystem storage) and on staging (S3 storage)


tortila:
[Verification form: uploaded files (green evidence) are not accessible after the request is submitted](https://trello.com/c/G4U92eKS/122-verification-form-uploaded-files-green-evidence-are-not-accessible-after-the-request-is-submitted)

PR #419, opened 2023-03-03
Add checks for empty formsets in form wizard

mrchrisadams:
This PR introduces checks at the formset level, to avoid cases where someone is able to work through the wizard form without adding location data or supporting evidence.

Previously, if we didn't have this check, using the django formset factory API, a user could sail through without adding the necessary info.

If we _did_ have the check though, it turned out that the formset errors we not surfaced onto the page, so users would be stuck not understanding why the step could not progress.

This PR introduces logic to add our usual defaults, but also to change the `errors` method logic on our multiform(s), so we display the correct error for the template to render.

Finally, this PR tweaks the error message so it's easier to understand in the template (we change the word 'form' to 'location')

### DIFF
--- a/apps/accounts/tests/test_provider_request.py
+++ b/apps/accounts/tests/test_provider_request.py
@@ -50,7 +50,7 @@ def wizard_form_org_location_data():
     """
     return {
         "provider_registration_view-current_step": "1",
-        "locations__1-TOTAL_FORMS": "0",
+        "locations__1-TOTAL_FORMS": "1",
         "locations__1-INITIAL_FORMS": "0",
         "locations__1-0-country": faker.country_code(),
         "locations__1-0-city": faker.city(),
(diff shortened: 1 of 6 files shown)


apps/accounts/tests/test_provider_request.py
```diff
@@ -50,7 +50,7 @@ def wizard_form_org_location_data():
     """
     return {
         "provider_registration_view-current_step": "1",
-        "locations__1-TOTAL_FORMS": "0",
+        "locations__1-TOTAL_FORMS": "1",
```
mrchrisadams, line 53: this was needed - it looks like the formset checks look for this value, even if the actual number of forms submitted might have correct in the fixture

mrchrisadams:
[Sanity check that all verification requests validate for at least one location](https://trello.com/c/wrkTZCly/108-sanity-check-that-all-verification-requests-validate-for-at-least-one-location)

PR #422, opened 2023-03-07
Use django-file-resubmit to retain files in case of ValidationError

tortila:
**The problem**:
When user uploads any files in the green evidence step of the verification request form, there are cases when the file goes missing:
- when user goes back to the green evidence step at some point, they see as if the file was missing (the file input is empty) and the validation logic (we expect exactly one: file or link) thinks that the file is missing too, leading to `ValidationError` when the form is re-submitted,
- when the validation error is raised (e.g. the step above occurs or user tries to upload a file and submit a link in the same form) - in that case files from *all* the forms in the formset are gone

**The solution**:
Installed a new library: [`django-file-resubmit`](https://github.com/un1t/django-file-resubmit), which provides a `ResubmitFileWidget`. Behind the curtains it stores all uploaded files to a cache, and:
- sets the value of the file input to the chosen file name (so that the user sees which file was uploaded)
- inserts a custom hidden input, which remembers which cached item should be used for the file.

The cache is configured to only keep the files for a maximum of 10 minutes, and a maximum of 300 items are allowed ([default value](https://docs.djangoproject.com/en/4.1/topics/cache/#cache-arguments)). 

**Gotchas**:
Currently there's no way to clear the file selection - the workaround is to delete the corresponding form from the formset.

This needs to be tested manually on production and staging to ensure that the user running the Django process has the correct permissions to the file cache location: `/tmp/file_resubmit/`.


tortila:
[Verification form: uploading documents don't persist if you go back in the form](https://trello.com/c/WFd3LJCL/97-verification-form-uploading-documents-dont-persist-if-you-go-back-in-the-form)

mrchrisadams (approved):
Thanks for adding this Oliwia - I'm really glad we got to the bottom of this one.

PR #434, opened 2023-04-12
Hs copy edits

hanopcan:


hanopcan:
Hi @mrchrisadams I want to merge this pull request in but the checks have failed. I'm not sure why. Would you be able to take a look? If it's straight forward for you to resolve would you go ahead and resolve, and then update staging with the latest code? Thanks :smile:

tortila:
@hanopcan I ran into the same issue with failing checks. It's a very peculiar situation: one of the dependencies (which we don't even utilize) was removed from the official package registry :exploding_head: What you can do to fix it: either wait until https://github.com/thegreenwebfoundation/admin-portal/pull/435 is merged and rebase your branch, or fix it directly in your branch: remove `codecov` from `Pipfile`, then run `pipenv lock`, then add and commit newly updated `Pipfile` and `Pipfile.lock`

Issue #438, opened 2023-04-18
[ENHANCEMENT]  Adding some support for carbon.txt in our detailed checkers and API involves us 


mrchrisadams:
Adding some support for carbon.txt in our detailed checkers and API involves a few changes:

- [x] **We need to make the extended lookup explicitly tell you when a lookup has been connected to a provider** via the green domains table.
- [x] **We need to update the API to support these carbon.txt lookups, even in `nocache` mode.** At present, a no-cache API check can delete the greendomain instance, )
- [x] **We need to update our carbon-txt preview page to allow folks to see how `Via` and DNS lookups are affecting which file is parsed** At present, we do not surface this - we just show matching providers for each domain.

PR #440, opened 2023-04-19
Add support for Via Header and DNS lookups in carbontxt preview

mrchrisadams:
This PR adds support for debugging carbontxt files.

When you add just an url, the parser preview now follows header `Via` information, as well as looking for DNS TXT records on a given domain.

This makes it possible to see where a chain of lookups will follow when testing your own carbon.txt file at a given url, as well as test which domains already ready return information for a given url.


mrchrisadams:
Add the previews showing Via and support for DNS TXT records

 https://github.com/thegreenwebfoundation/admin-portal/issues/438

PR #441, opened 2023-04-19
Key our results in carbontxt previews by domain

mrchrisadams:
This changes the carbontxt preview to screen to show the results of each domain lookup keyed to the domain used. This makes it clearer what provider we found when doing a lookup


mrchrisadams:
I'm closing this in favour of merging in https://github.com/thegreenwebfoundation/admin-portal/pull/442, which has the correct issue in the name  for future reference.

PR #448, opened 2023-04-25
Record the verification request approval date

tortila:
### Scope:
- a new, non-editable datetime field on the `ProviderRequest` model + migration
- the field is populated with current datetime when the verification request is approved
- modified existing test to verify the behavior
- if there's an approval date, it's displayed on the detail view of the `ProviderRequest`
- additionally I fixed a small bug where the success message upon submitting a new request was rendered twice (once from detail.html and once from base.html template)

### To do
Apply migrations manually after the deployment is done


apps/accounts/models/provider_request.py
```diff
@@ -50,6 +50,7 @@ class ProviderRequest(TimeStampedModel):
     created_by = models.ForeignKey(
         settings.AUTH_USER_MODEL, on_delete=models.CASCADE, null=True
     )
+    approved = models.DateTimeField(editable=False, blank=True, null=True)
```
mrchrisadams, line 53: We currently show the approved status, but in the admin view, I couldn't see the time of approval listed in when visiting `/admin/accounts/providerrequest/5/change/`.

My guess is that it would make sense to add it near the status part of the admin view, as they seem linked to me.

mrchrisadams:
Hi @tortila, thanks for making these changes. there's a just a couple of small tweaks I've requested, and then I think it's good to merge in.

Can you message me on zulip to talk about running the migration?

mrchrisadams (commented):
Two small changes needed, but otherwise this is good to go 👍

PR #463, opened 2023-05-05
Only display approved verification requests on the provider portal landing page

tortila:
Addresses feedback from https://trello.com/c/wpIzRRsg/165-edit-provider-portal-copy-to-make-the-difference-between-requests-and-providers-clearer.

Scope:
- exclude approved requests from rendering on the provider portal home page,
- add a test to verify the new logic

### DIFF
--- a/apps/accounts/tests/test_provider_portal.py
+++ b/apps/accounts/tests/test_provider_portal.py
@@ -1,0 +1,33 @@
+from ..views import ProviderPortalHomeView
+from ..models import ProviderRequestStatus
+from conftest import ProviderRequestFactory, ProviderRequestLocationFactory
+from django.test import RequestFactory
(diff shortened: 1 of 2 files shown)


mrchrisadams (commented):
Hi Oliwia, the logic looks fine here - my only feedback is that I couldn't immediately link what we being requested in the trello card to the test method name, and it took a bit of close reading to see what it was trying to show.

If you can make the intention clearer either in the comments or method name, then this looks great to me.

tortila:
Thanks for your feedback @mrchrisadams, I addressed it in https://github.com/thegreenwebfoundation/admin-portal/pull/463/commits/04ae6539e7eed1da6e28bdb26e2b730558538756

PR #467, opened 2023-05-08
Add migrations, mgmt command and admin updates for migrating service names

mrchrisadams:
This PR introduces:

1. **new, separate models for Services and our intermediate model Provider services** - this removes the need for the confusing 'taggit' app in the admin, and allows us to extend the Service model


tortila (approved):
LGTM :+1: 
We reviewed this together and discussed the outstanding action items for releasing this change (snapshot of the db before the migrations and management commands are run, cleaning `ProviderRequest` data on staging).

PR #474, opened 2023-05-25
Add mail-to-board for submitted verification requests

mrchrisadams:
This PR introduces the use of the mail-to-board features of Trello, so that when someone completes a verificaiton request, an email is sent to the mail-to-board address we use for a trello board we use to track the status of verification requests.

This makes it possible to introduce a clearer audit trail, and use all of trello's workflow features for tracking who is allocated to a request, what its status is and so on, instead of trying to implement them in Django ourselves.


greenweb/settings/testing.py
```diff
@@ -22,3 +22,4 @@
         # "dramatiq.middleware.Retries",
     ],
 }
+TRELLO_REGISTRATION_EMAIL_TO_BOARD_ADDRESS = "mail-to-board@trello.com"
```
tortila, line 25: Question: for testing environment, should we use an email address that is not external? E.g. `mail-to-board@localhost`. I'm not sure if the current setup will successfully prevent sending external emails

tortila (approved):
I added a comment with a question, but overall the changes look good :+1:

PR #479, opened 2023-06-14
Add commands for easier tracking of data quality

mrchrisadams:
This PR introduces a management command, and a convenience function for looking at coming expiry dates for providers.

It seemed like something we'd want to project forwards so we know when we can predict expected workload for our support desk

### DIFF
--- a/apps/accounts/models/hosting.py
+++ b/apps/accounts/models/hosting.py
@@ -1,6 +1,8 @@
 import logging
 
 import secrets
+import typing
+import datetime
 from django.db import models
 from django.conf import settings
 from django_countries.fields import CountryField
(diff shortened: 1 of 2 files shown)


apps/accounts/models/hosting.py
```diff
@@ -284,9 +286,26 @@ def shared_secret(self) -> str:
...
+    @property
     def public_supporting_evidence(self):
+        """
+        Return the supporting evidence that has explictly been marked as public
+        by the users uploading it to the database
+        """
         return self.supporting_documents.filter(public=True)
 
+    @property
+    def evidence_expiry_date(self) -> typing.List[datetime.date]:
```
tortila, line 298: From the method body I assume the expected return type is `Optional[datetime.date]`

tortila (approved):
Looks good :+1: Left a minor comment on the typing annotation

tortila (approved):

PR #480, opened 2023-06-14
Add staff actions for rejecting and removing verification requests

mrchrisadams:
This PR adds the ability to reject and remove


tortila (commented):
Looks good :+1: Could you add a test (or modify an existing one) that will cover the case of filtering out removed requests from the provider portal home page?

mrchrisadams:
hi @tortila - would you cast a quick eye over this test, and if you're ok with it merge this change in?

tortila (approved):

Issue #335, opened 2022-11-16
Users should be able to manage more than 1 hosting provider


tortila:
Currently there's a one-to-many relationship between hosting providers and users (via foreign key from User to Hostingprovider). This allows a single user to manage a single Hostingprovider, and multiple users can manage the same Hostingprovider. Now we want to enable users to manage multiple hosting provider objects.

Depending on whether we want to continue allowing multiple users to manage a single hosting provider, we should change this relationship to many-to-one or many-to-many

tortila:
![](https://github.trello.services/images/mini-trello-icon.png) [Users should be able to manage more than 1 hosting provider](https://trello.com/c/OeZn4Fhw/42-users-should-be-able-to-manage-more-than-1-hosting-provider)

mrchrisadams:
I'm adding some links

This was the blog post about a well through through way to a grant access to objects give as a really good overview of the theory we might want to be aware of:
https://tailscale.com/blog/rbac-like-it-was-meant-to-be/

This talk here from djangocon is probably the best talk I can find to help understand common patterns for providing rich access control and object level permissions with django - it'd be really good to review before the planning session next week for this.
https://www.youtube.com/watch?v=svTt9F7MnDk

tortila:
I started implementing object-level permissions using `django-guardian` and so far it looks like it covers all of our needs:
- build-in admin integration on the object change view (we can manage who has access to a hosting provider X on their admin page)
- users as well as groups are supported: https://django-guardian.readthedocs.io/en/stable/userguide/assign.html#for-group - this can be utilised in the future to implement a concept of organisations, covering multiple hosting provider objects, based on groups, and managing permissions on organisation level (FYI @mrchrisadams)
- our custom user admin needs to be extended to allow managing object-level permissions for a given user

PR #481, opened 2023-06-14
Object-level permissions for hosting providers and data centers

tortila:
*This is a draft pull request - documentation is still missing!*
Closes https://github.com/thegreenwebfoundation/admin-portal/issues/335.

This pull request introduces object-level permission system, which can be used to assign arbitrary number of users a permission to manage (through admin or API) an arbitrary number of hosting providers and data centers.

**Scope of changes:**
- introduced a library [`django-guardian`](https://django-guardian.readthedocs.io/en/stable/) to provide a back-end for object-level permissions,
- introduced custom permissions: `manage_provider` for `Hostingprovider` model and `manage_datacenter` for `Datacenter` model. The permissions are defined as singletons for re-using in the app (new module `permissions.py`).
- extended `GuardedModelAdminMixin` provided by `django-guardian` so that only admins can access permission management screens, and only custom model permissions are displayed (excluding default Django permissions),
- created a data migration to assign global permissions (i.e. to all objects in the database) for the admin group,
- implemented data migrations so that existing relationships between Users and `Hostingprovider`s, and Users and `Datacenter`s are expressed as object-level permissions,
- new fields: `Hostingprovider.created_by` and `Datacenter.created_by`, data populated for existing records based on the User.hostingprovider reference.
- removed FK reference from `User` to `Hostingprovider`,
- Both models, `Hostingprovider` and `Datacenter` have now helper properties: `users` and `users_explicit_perms` that each returns a queryset of related users based on defined permissions,
- `User` model has now helper properties: `hosting_providers` and `data_centers` that each returns a queryset of related objects based on defined permissions,
- admins for `Hostingprovider`, `Datacenter` and `User` models now include a list of related objects based on defined permissions,
- provider portal home page lists all available hosting providers for a user based on defined permissions,
- IP ranges / ASN API was updated to return objects across all providers that the requesting user has access to,
- added new tests to cover the above & updated existing ones.

**To do:**
- modify the `ProviderSharedSecretView` API, so that the provider is explicitly identified (this will be handled in a separate task),
- add documentation,
- deploy to staging & verify the functionality


apps/accounts/migrations/0055_populate_existing_permissions.py
```diff
@@ -0,0 +1,127 @@
...
+            created_by__isnull=False
+        )
+    ]
+    # Get current models to retrieve the objects for assigning permissions.
+    # This is dangerous! Future versions of django-guardian might change the schema
+    # and this migration might fail when ran from scratch. If this happens:
+    # - pin down the version of django-guardian to the latest working version
+    #   so that migrations work for tests
+    # - for production / staging it's a non-issue since we always keep snapshots around
+    #   and don't run older migrations
```
mrchrisadams, line 89: Would you mind adding a link to the issue in question where we discovered this and found the workaround? I can see this being a pain to lookup again, as took a while for us to understand before.

https://github.com/django-guardian/django-guardian/issues/281

apps/accounts/migrations/0055_populate_existing_permissions.py
```diff
@@ -0,0 +1,127 @@
...
+    Hostingprovider and Datacenter objects.
+
+    Parameter `manage_func` is either: assign_perm or remove_perm from guardian.shortcuts
+    """
+    db_alias = schema_editor.connection.alias
+
+    hp_perm, dc_perm = _get_managed_perms(apps, db_alias)
+
+    # to query the objects eligible for assigning permissions we need the historical model
+    # so that we can use queries based on fields present at the moment of the migration
```
mrchrisadams, line 66: Woud you please add a link to some explanatory information about historical models in this?

I think it's a pretty advanced topic, and future me would really appreciate knowing there was documentatino available to help understand why models don't have the properties I was expecting, or what happend to the methods I mightt otherwise expect to call.

I found this, but there maybe others of value too.

https://docs.djangoproject.com/en/4.2/topics/migrations/#historical-models

https://docs.djangoproject.com/en/4.2/topics/migrations/#historical-models

apps/accounts/models/hosting.py
```diff
@@ -47,10 +48,40 @@ class Datacenter(models.Model):
...
     )
-    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    created_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True
+    )
     virtual = models.BooleanField()
     website = models.CharField(max_length=255)
 
+    @property
+    def users(self) -> models.QuerySet["User"]:
```
mrchrisadams, line 58: Thank you for adding the type hints 💚 

It would be wonderful if this become the norm, but until we're using python 3.11 or 3.12 across the board, it's a "nice to see", not a hard requirement.

apps/accounts/models/hosting.py
```diff
@@ -263,6 +300,30 @@ class Hostingprovider(models.Model):
...
+
+    @property
+    def users_explicit_perms(self) -> models.QuerySet["User"]:
+        """
+        Returns a QuerySet of all Users that have *explicit* permissions to manage this Hostingprovider,
+        not taking into consideration:
+            - group membership
+            - superuser status
+        """
+        return get_users_with_perms(
```
mrchrisadams, line 320: Would you please add a line in the code commentsd about why we have this function?

It's not really needed in the docstring, but it's not obvious that without it, the admin screens become a horror show of massive listings of providers a user has access to, and knowing this would be helpful for understanding why we have two functions that seem to do similar things.

apps/accounts/models/test_models.py
```diff
@@ -46,21 +46,18 @@ class TestHostingProvider:
             "compensatie",
         ),
     )
+    @pytest.mark.django_db
```
mrchrisadams, line 49: I see you using this decorator over paassing `db` into the function, and we have a mix of use throughout our test suite.

I'm open to us settling on either, but I'm not sure about the reasoning either way. 

Is there a reason for your preference here? I can see some value in splitting it out as a decorator, as it lets us decorate an entire test class, that needs to hit the db, but all the new tests seem to use function name as a namespacing strategy, so it wasn't so clear to me why we'd use one over the other.

apps/accounts/models/user.py
```diff
@@ -86,6 +79,63 @@ def __str__(self):
...
+        """
+        Returns a QuerySet of all Datacenters that the User has permissions to manage.
+
+        This method is useful to verify whether a user has permission to manage the datacenter.
+
+        """
+        return get_objects_for_user(self, str(manage_datacenter))
+
+    @property
+    def data_centers_explicit_perms(self) -> models.QuerySet[Datacenter]:
```
mrchrisadams, line 121: As above with the comments about why we have an explicit method. I suspect it won't be clear why we have to similar methods in a few months, and it would be nice to be kind to our future selves.

apps/accounts/permissions.py
```diff
@@ -0,0 +1,27 @@
+from dataclasses import dataclass
+
+
+"""
+Definitions for object-level permissions for the `accounts` app.
+"""
+
+
+@dataclass
+class Permission:
```
mrchrisadams, line 10: I am a bit confused about the difference between this and [the Django Permission model](https://docs.djangoproject.com/en/4.1/ref/contrib/auth/#permission-model), and why the dataclass appears to be being used instead of the normal Permission model.

Would you add a comment or explain what's going on here? I wouldn't feel that confident explaining this part to someone else right now.


We have this dataclass to

apps/accounts/tests/test_admin.py
```diff
@@ -149,6 +146,22 @@ def test_get_inlines_non_staff(
...
         assert "Associated hosting providers" not in fieldset_names
 
+    def test_list_of_users_who_can_manage_dc_is_displayed(
+        self, db, sample_hoster_user, greenweb_staff_user, datacenter
+    ):
+        # given: 2 distinct users have explicit permissions to manage the datacenter
+        assign_perm(str(manage_datacenter), sample_hoster_user, datacenter)
+        assign_perm(str(manage_datacenter), greenweb_staff_user, datacenter)
+
+        # when the datacenter admin page is accessed by one of the users
```
mrchrisadams, line 156: This is looks like a really nifty way to test Django Admin functionality without needing to faff with creating a client, logging them in and so on (which is the way I would have thought I had to test it)

Is there a blog post or page you might point to explain this in more detail?

It's ok if not, but I this looks like the first instance iv'e seen of instantiating the admin site like this.

apps/accounts/tests/test_admin.py
```diff
@@ -360,60 +373,98 @@ def test_archived_providers_hidden_by_seen_with_override_params(
...
-        }
+    def test_admins_can_access_permissions_management(
+        self, db, greenweb_staff_user, hosting_provider, sample_hoster_user
+    ):
+        hosting_provider.save()
+        # when the provider admin page is accessed by admin user
+        provider_admin = ac_admin.HostingAdmin(
+            ac_models.Hostingprovider, admin_site.greenweb_admin
+        )
+        request = MagicMock()
```
mrchrisadams, line 413: We've previously used [django.test.RequestFactory](https://docs.djangoproject.com/en/4.1/topics/testing/advanced/#django.test.RequestFactory) to create requests for using in testing like this, as it's well documented in the django docs. 

Is there specific reason for using MagicMock? If not, would you mind using the django docs? The docs are slightly easier to read, and finding good docs about mocks can be a pain.



If not, would you mind using the RequestFactory?

apps/greencheck/api/views.py
```diff
@@ -97,15 +97,15 @@ def get(self, request, ip_to_check=None, format=None):
...
     serializer_class = ProviderSharedSecretSerializer
     authentication_classes = [SessionAuthentication, BasicAuthentication]
-    permission_classes = [BelongsToHostingProvider]
 
     @swagger_auto_schema(tags=["Provider Shared Secret"])
     def get(self, request, format=None):
-        try:
-            provider = request.user.hostingprovider
-        except AttributeError:
+        # implicitly choose a first provider that the user has permissions to manage
```
mrchrisadams, line 106: Would you please add a line saying something like `TODO: revisit when updating API with revised permissions in production`? We'll definitely need to come back to this, but I want to make  it clear to anyone looking to integrate with us that we are planning to support multiple providers in future.

apps/greencheck/tests/test_admin.py
```diff
@@ -81,9 +77,12 @@ def test_get_queryset_staff(
         assert gip in qs
 
     def test_get_queryset_non_staff(
-        self, db, rf, hosting_provider_with_sample_user, green_ip,
+        self,
+        db,
```
mrchrisadams, line 81: If we're going to use the `django_db` mark in pytest passinf in `db` as a fixture, would you mind making the change here too so it's consistent in this PR?

apps/greencheck/tests/test_api_asn.py
```diff
@@ -195,12 +232,13 @@ def test_can_only_destroy_asns_for_own_hosting_provider(
...
          to create or destroy ASNs or IP Ranges for the provider
         """
+        green_asn.hostingprovider = hosting_provider_with_sample_user
         green_asn.save()
         assert gc_models.GreencheckASN.objects.filter(asn=12345).count() == 1
 
         rf = APIRequestFactory()
         url_path = reverse("asn-detail", kwargs={"pk": green_asn.id})
-        hosting_provider_with_sample_user.user_set.first()
+        # hosting_provider_with_sample_user.users.first()
```
mrchrisadams, line 241: It doesn't look like this code is doing anything anymore (if it ever did). Would you mind deleting it if so?

mrchrisadams (approved):
Hi Oliwia! This looks good! 

I see no issues with the logic - my feedback is mainly about style and consistency.

There are are a couple of places where I couldn't understand why we might use one class created in this PR over a built in one (like with Permissions), but otherwise I'm happy.

We discussed the precautions we might take before merging this into production, and I think I'll need to speak to to you in off github and in Zulip about how we sequence deployment and running the migration.

Thanks again for this picking up this massive task.

apps/accounts/models/test_models.py
tortila: It's only a matter of personal preference - I prefer to have a decorator over adding a function argument. But frankly I didn't think much of it, and I don't think it's an impactful decision for the codebase. If you have a strong preference let me know, I can change this to a function parameter

apps/accounts/tests/test_admin.py
tortila: I literally copied it from the test above... which you added/modified :grin: 
https://github.com/thegreenwebfoundation/admin-portal/pull/481/files#diff-030e218c1ff710051a2f710b11bc6290c4d32c8f15e6f4c1b6cdcce9028c1877R125-R127
But I agree it's very handy!

apps/accounts/tests/test_admin.py
tortila: I this specific test case, I need to pass any object to the view function that will have a `user` attribute set correctly, and the behavior I'm testing is only depending on this parameter - that's why I used `MagicMock`. It would be inconvenient to use `RequestFactory` in that specific test case, because I'm covering 3 different views

apps/greencheck/api/views.py
tortila: There's a TODO item a couple of lines above: https://github.com/thegreenwebfoundation/admin-portal/pull/481/files#diff-1509052d177b4feeced011bc5b0c2a26c9837bf7f33abfb09beb26f6b2272640R100

apps/greencheck/tests/test_admin.py
tortila: It looks like we use both version across our test cases. In each new test I went with the variant that is already used in other tests in the same module. Do you want to take the decision now about this and update all occurrences? I don't see it as important, but let me know if you think otherwise

apps/greencheck/tests/test_api_asn.py
tortila: Good catch!

apps/greencheck/tests/test_admin.py
mrchrisadams: Nah, not's not a priority.

Let's keep this as is and create an issue to review this where make a recorded decision and then apply it to all future PRs instead. This PR is already ginormous.

apps/accounts/models/test_models.py
mrchrisadams: See below - we can punt this til later. No decisions or changes needed for right now.

PR #483, opened 2023-06-15
Update approval process to annotate providers and check for non-hosting orgs

mrchrisadams:
This PR introduces support organisations that do not offer hosted services, when sending verification requests, as well as defaulting to showing providers in the directory by default after they have been approved by staff.


mrchrisadams:
Merging this in, becuase @tortila just stood next to me in person, discussed it and gave a physical thumbs up, approving in meatspace.

PR #489, opened 2023-07-11
Draft - update dependencies to support building development environment on OS X

mrchrisadams:
This PR tightens up the specified version of `mysqlclient` that was causing failing builds on OS X, which meant working without gitpod locally was impossible.

It also moves the `rich` logging library into the `packages` group. It's useful in management commands, as well as logging generally.


mrchrisadams:
At least.

With pipenv, I think you need to be really careful when setting up a virtualenv. This didn't give me the virtual env using python 3.8.10, no matter what I did with switching local pythons, and even making it explicit in the pipfile doesn't seemt to work.

```
PIPENV_VENV_IN_PROJECT=True python -m pipenv lock --dev

I had to do this when creating the virtual environment instead:

```
PIPENV_VENV_IN_PROJECT=True python -m pipenv lock --dev --clear --python 3.8.10
PIPENV_VENV_IN_PROJECT=True python -m pipenv install --dev --clear --python 3.8.10
```

mrchrisadams:
This is no longer needed - we addressed the problematic mysqlclient dependency issue in a recent PR instead linked below:

https://github.com/thegreenwebfoundation/admin-portal/pull/496

PR #490, opened 2023-07-12
Make the error message clear when users are accessing another user admin page

tortila:
As a regular user, when I go to a hosting provider change view in the admin panel, I can see other users who can manage that provider. If I click on one of the usernames that is not mine, I'm getting redirected to the main admin page and I see a message: "User with ID XXX doesn't exist. Perhaps it was deleted?". The default message is somewhat confusing and indicates that the user could have been deleted. This is because for non-admin users we filter the queryset to only include their own object - which makes Django deduct that the user with given ID does not exist.

### DIFF
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -1,9 +1,11 @@
+from typing import Optional
 from django.db import models
 from django.core.mail import send_mail
 from django.core.exceptions import PermissionDenied
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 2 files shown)


tortila:
[User permissions: if I click on another user assigned to the same hosting provider as me, what should I see about them?](https://trello.com/c/YJua5v34/210-user-permissions-if-i-click-on-another-user-assigned-to-the-same-hosting-provider-as-me-what-should-i-see-about-them)

PR #492, opened 2023-07-13
Add sequence flow for main API greencheck traffic

mrchrisadams:


mrchrisadams:
Hey @tortila  can you please cast an eye over this an merge in when done? It contains the sequence diagrams of the most common flows of data we need to think about.

docs/understanding-the flow-of-a-greencheck.md
```diff
@@ -0,0 +1,66 @@
...
+```{mermaid}
+sequenceDiagram
+
+    Django Dramatiq->>+RabbitMQ: Check for any domains to log
+    RabbitMQ->>+Django Dramatiq: Return domain to log
+    Django Dramatiq->>+Database: Log domain to greencheck table
+    Django Dramatiq->>+Database: Update greendomain tables
+```
+
+We can scale the two of these independently, depending on the traffic we are receiving, and RabbitMQ here acts like a buffer.
```
tortila, line 45: > We can scale the two of these independently

which two things are implied here?

docs/understanding-the flow-of-a-greencheck.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Carrying out a slow full network look up
```
tortila, line 47: When is the slow check used instead of the quick lookup?

docs/understanding-the flow-of-a-greencheck.md
```diff
@@ -0,0 +1,66 @@
...
+    Nginx->>+Django Web: Forward to a django web process
+    Django Web->>+Database: Look up domain
+    Database->>-Django Web: Return lookup result
+    Django Web->>+Nginx: Return rendered result
+
+    Nginx->>+Browser client: Present domain result
+    Django Web->>+RabbitMQ: Queue domain for logging
+```
+
+In most cases we try to find a result we can return quickly, and check in a local cache table, described by the GreenDomain model. Assumign we find a result in the database, we put the checked domain on a Rabbit MQ queue, so that a separate worker process can update to the cache to record the time of the check, and so it can write the check result to a logging table.
```
tortila, line 24: Small typo: Assumign -> Assuming

docs/understanding-the flow-of-a-greencheck.md
mrchrisadams: I've updated the docs to make this clearer 👍

mrchrisadams:
Merging in after an in-person convo with Oliwia

PR #493, opened 2023-07-19
Use custom `change_form` template for Hostingprovider & Datacenter admin

tortila:
This PR fixes the issues [described in this Trello card](https://trello.com/c/Wu8rqMlv/213-post-migration-cleanup-add-support-for-labels-back-in). 

The issues were introduced when I chose to extend the default change_form admin template for Hostingprovider and Datacenter models as a part of introducing object level permissions for them. I now reverted that to extend the custom template. On top of that, I made some small modifications so that template logic for displaying authorised users is moved to the corresponding admin.


tortila:
[Post migration cleanup: add support for labels back in](https://trello.com/c/Wu8rqMlv/213-post-migration-cleanup-add-support-for-labels-back-in)

mrchrisadams (approved):
The only caveat was the multi select thing for widget . I think the widget we had in use in production is called `django.forms.SelectMultiple`:

https://docs.djangoproject.com/en/4.2/ref/forms/widgets/#django.forms.SelectMultiple

PR #499, opened 2023-07-27
Add regrouping of providers in directory

mrchrisadams:
This PR introduces our list of providers grouped by country, as outlined in the Trello card below:

https://trello.com/c/zfA4swmo/183-directory-output-list-of-providers-by-country

It also introduces a couple of simple tests.

The first is a sanity check that requests sent to `/directory` work.

The second is a check that when we have our list of providers we are ordering them by country name. We use django's built in [#regroup](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#regroup) template tag to reorder the providers in the template, because this makes it easier to understand what's going on when Han is working at the template layer, and because implementing out own regrouping logic seemed to replicate what django offers us already. 

The main consideration was that becuase we use a specific country model from django-countries, we can't rely on a ORM query to order our providers by country name using the database - this is why we use sorted, and our own sorting function.

WIthout this, Germany (DE) is placed ahead of Denmark (DK) in our grouping, leading to a confusing ordering of providers of results.

It's likely that we'll want to make further changes to the appearance in the directory, but given we have card 182 about showing the details about the providers in more detail, I figure it would be easier to address this one first, and then make a separata PR once we've made a few more decisions about what we show in the listings there.

https://trello.com/c/f8TkBAAE/182-directory-output-additional-details-on-list-of-providers

### DIFF
--- a/apps/greencheck/tests/test_directory.py
+++ b/apps/greencheck/tests/test_directory.py
@@ -1,4 +1,3 @@
-from django.test import RequestFactory
 from django.urls import reverse
 from waffle.testutils import override_flag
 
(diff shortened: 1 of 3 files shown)


tortila (approved):
Looks good :+1:

PR #503, opened 2023-08-01
Hs directory mvp styling

hanopcan:


apps/theme/templates/greencheck/partials/directory_results.html
```diff
@@ -0,0 +1,114 @@
...
+					{% if not obj.public_supporting_evidence and not obj.description %}
+						<details class="">
+							<summary class="text-right text-sm text-neutral-400">Help improve this listing</summary>
+
+							<p class="max-w-xs md:max-w-md text-sm text-justify ml-auto mr-0 mt-2 text-neutral-400">
+							Once a hosting provider gets verified with us, we ask them to keep
+							their listing up to date and refresh their evidence at least annually.</p>
+
+							<p class="max-w-xs md:max-w-md text-sm text-justify ml-auto mr-0 mt-2 text-neutral-400">
+							<span class="text-neutral-500">Are you a customer of this provider?</span> We find
```
mrchrisadams, line 99: We have a sample emails page, linked below - would you mind linking to that? It might be a separate card to add a sample email to ask for this change, but having at least some copy and pastable content on a url we can send people to would provide an action we can see if people are taking.

https://www.thegreenwebfoundation.org/sample-emails/

mrchrisadams (commented):
Hi @hanopcan thanks for this - there's a couple of fairly small changes, but otherwise this looks good. Thanks again 👍

apps/theme/templates/greencheck/directory_index.html
```diff
@@ -3,23 +3,62 @@
...
+				</p>
+			</details>
+
         </section>
     </div>
-    <section class="main-content container mx-auto">
+
+    <section class="main-content container mx-auto md:mb-8 mt-4 lg:mt-6">
         {% if not ordered_results %}
-            {% include "greencheck/partials/_no_directory_results.html" %}
```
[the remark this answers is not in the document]
hanopcan: Thanks @mrchrisadams - I didn't know about that convention and I can see how useful it would be. I'll change things back. Thanks for explaining

apps/theme/templates/greencheck/partials/directory_results.html
hanopcan: Yes happy to add that in. I will draft a sample email on that page at the same time to cover this case so we have at least something that can be improved later.

hanopcan: Sample email made on the page, and the other emails and content in general given a bit of love and attention. I think that closes this card off for now.

PR #508, opened 2023-08-24
Edit a verification request

tortila:
This PR aims to enable editing of submitted verification requests by their authors, using the same multi-step wizard form.

**Scope of changes:**
- New URL path to the edit view: `/requests/{id}/edit` which is only accessible for admins and authors, and only when the request is open. We re-use the same WizardView but this time pass an `instance_dict` as an argument, which contains model instances as expected by consecutive steps.
- All forms used by the wizard are now either `ModelForm` or `ModelFormSet`s. This allows to inject either an `instance` of `queryset` with existing objects and delegate handling the updating/deletion of objects to Django magic.
- Status "Open" now renamed to "More info required", added a relevant migration.
- New button on the detail page of a verification request: "Update this request" will be visible to request's author and admins when the request has status "More info required". The button redirects to the edit view.
- Some workarounds for `MultiModelForm`s and `ConvenientFormSet`s for the whole machinery to work properly (all workarounds documented in the code with examples).
- New action added to the admin panel: requesting more information about a `ProviderRequest` object will mark its status to OPEN.
- Clean up models for representing consent (last step of the verification request form): now consent fields are a part of the `ProviderRequest` model and not a separate one. Added a relevant migration and changed the model forms.
- Small change in the CSS: added a rule to always mark elements with attribute "hidden" as `display :none`. This is a workaround so that evidence rows can be deleted, this didn't work before because the flexbox overrides the "hidden" attribute behavior.
- Added some tests, including `test_editing_pr_updates_original_submission` which covers an end-to-end journey of updating a verification requests.

**Out of scope:**
- attaching notes/comments by staff to opened verification requests,
- automatic email communication with the author in case a verification request is re-opened.

**After merging:**
Remember to run the migrations!


mrchrisadams:
hi @tortila  thanks for taking me through this on the call just now - I feel like I have a good understanding of the changes now, and the last few bits before we merge it in. Have a nice week away :)

mrchrisadams (approved):
Oliwia and I reviewed this code synchronously on a video call, and again in person. I'm happy to merge this in.

PR #517, opened 2023-09-19
Add support for sentry tracing

mrchrisadams:
This PR allows us to set Sentry's tracing on our apps on a per machine basis, so we can get an idea of where performance bottlenecks lie. 

Previously we had api.TGWF and admin.TGWF running on the same server, but since expanding to more machines, this allows us to serve admin.TGWF (which receives like.. 1000x less traffic) with sentry tracing on the app3 server, where we could run the tracing without blowing through our quota, and api.TGWF on the other machines, where we do not.

This also supports sampling on the API servers, but because admin.TGWF still gets some API traffic, this is currently assumed to be less of a priority.

### DIFF
--- a/ansible/templates/run_gunicorn.sh.j2
+++ b/ansible/templates/run_gunicorn.sh.j2
@@ -1,4 +1,4 @@
 # supervisor can only control processes it started itself.
 # So we need to use exec to replace the parent shell script process
 # that starts pipenv
-exec python -m pipenv run gunicorn greenweb.wsgi -b {{ internal_ip }}:$PORT -t 300 -c gunicorn.conf.py --statsd-host=10.0.0.2:9125 --statsd-prefix=member.app
+exec python -m pipenv run gunicorn greenweb.wsgi -b {{ internal_ip }}:{{ gunicorn_port }} -t 300 -c gunicorn.conf.py --statsd-host=10.0.0.2:9125 --statsd-prefix=member.app
--- a/ansible/templates/supervisor.gunicorn.conf.j2
+++ b/ansible/templates/supervisor.gunicorn.conf.j2
@@ -1,19 +1,15 @@
 # {{ ansible_managed }}
 # Last run: {{ template_run_date }}
-
 [supervisord]
 environment=LC_ALL='en_US.UTF-8',LANG='en_US.UTF-8'
-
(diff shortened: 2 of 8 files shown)


ansible/templates/supervisor.gunicorn.conf.j2
```diff
@@ -1,19 +1,15 @@
...
-
 [supervisord]
 environment=LC_ALL='en_US.UTF-8',LANG='en_US.UTF-8'
-
 [program:{{ supervisor_gunicorn_app }}]
 directory=/var/www/{{ tgwf_domain_name }}.thegreenwebfoundation.org/current/
 numprocs=1
 command=bash ./run_gunicorn.sh
 process_name=%(process_num)02d
-environment=PORT=90%(process_num)02d
```
mrchrisadams, line 12: We don't scale with supervisor - gunicorn has good support for forking subprocesses and threads, so we do it there instead.

ansible/templates/run_gunicorn.sh.j2
```diff
@@ -1,4 +1,4 @@
 # supervisor can only control processes it started itself.
 # So we need to use exec to replace the parent shell script process
 # that starts pipenv
-exec python -m pipenv run gunicorn greenweb.wsgi -b {{ internal_ip }}:$PORT -t 300 -c gunicorn.conf.py --statsd-host=10.0.0.2:9125 --statsd-prefix=member.app
+exec python -m pipenv run gunicorn greenweb.wsgi -b {{ internal_ip }}:{{ gunicorn_port }} -t 300 -c gunicorn.conf.py --statsd-host=10.0.0.2:9125 --statsd-prefix=member.app
```
mrchrisadams, line 4: We set the port explicitly because this allows us to run production and staging apps on the same machine, without port clashes

greenweb/settings/production.py
```diff
@@ -54,7 +54,7 @@
         # set our identifying credentials
         dsn=sentry_dsn,
         # Set traces_sample_rate.
-        traces_sample_rate=sentry_sample_rate,
+        traces_sample_rate=float(sentry_sample_rate),
```
mrchrisadams, line 57: Sentry seems to expect a float, not a string, so we coerce accordingly.

PR #518, opened 2023-09-21
Downgrade Django Autocomplete Light explaining why

mrchrisadams:
This resolves the issue we've seen in the django staff admin - previously we had the handy tag auto complete used by staff, OR the toggles for listing services a provider offers but not both.


mrchrisadams:
[Unable to select/change services in backend](https://trello.com/c/f9RZhGM2/270-unable-to-select-change-services-in-backend)

PR #519, opened 2023-09-21
Remove provider change form template overrides

mrchrisadams:
This cleans up the template to remove all our overrides we were relying on before, that updating dependencies has made unncessary.


mrchrisadams:
[Unable to select/change services in backend](https://trello.com/c/f9RZhGM2/270-unable-to-select-change-services-in-backend)

PR #522, opened 2023-09-27
Edit hosting provider using the multi-step form

tortila:
**Scope:**
- new "Update listing" button on the provider portal homepage is visible for listed hosting providers. This is only displayed when a waffle flag `edit_provider` is enabled.
- new URL path: `/providers/{provider_id}/edit/` is configured with a WizardView to display multi-step form with initial data filled in using data from existing `Hostingprovider` object. Submitting the form results in creating a new `ProviderRequest` object with the field `provider` set to `provider_id`
- new logic for the `approve()` method on the `ProviderRequest`: if the value of the `provider` field points to an existing provider, update that one, otherwise create a new one. When existing Hostingprovider is being updated, all its related objects are deleted and then re-created

**Bug: re-submitting initial data without any changes. - RESOLVED**
This one is tricky - ModelFormSets with `initial` data passed but without `queryset` argument provided behave in a strange way:
- `formset.initial_forms` is empty, `formset.initial_extra` contains the expected initial forms
- `formset.save()` for some reason does not create new instances as expected, possibly additional logic is needed to handle `initial_extra` forms
- submitted initial forms are treated as empty. At first I thought that this issue only affects validation logic for minimum number of forms: when `validate_min` is set to True in the `modelformset_factory` method. This can be worked around by setting this argument to `False` and adding a custom validation logic. But the problem is bigger, because:
- saving instances for initial data does not work - either no instances are returned by `formset.save()` or the instances returned are not valid. 

_this awkward resubmitting initial data bug appears to have been resolved now. We still need to check about how uploads are treated, but the hairy problemwe were strugglign with has a fairly straightforward solution now._

**After deployment:**
- run migrations


mrchrisadams:
We've just discussed this in person - the problem we come across is when a form is showing as _not changed_. If we can find a way to _always introduce a change_, like a hidden field with a timestamp or something, then that might allow us to bypass the "form not changed" issue.

mrchrisadams:
I've added a question in the django forums, and in the process, I think I've got more of clue about what we need to do to fix the unchanged data bug:

https://forum.djangoproject.com/t/is-there-way-to-override-changed-data-in-django-forms/24263/2

mrchrisadams:
Ok, some more thinking about this, and looking through the code.

The crux of the data is that we have forms reporting themselves as unchanged, and in the default formset handling code, is a form that is marked as unchanged is ignored.

We don't want that in our case, and after experimenting, I think a clean solution that doesn't involve diving too deeply into hairy formset handling logic is to just modify the forms themselves so they always return `True` when `has_changed()` is called.

```python

class IpRangeForm(forms.ModelForm):
    start = forms.GenericIPAddressField()
    end = forms.GenericIPAddressField()

    def has_changed(self):
        """
        Override our default behaviour defined in BaseForm for use in the wizard        
        """
        return True
```

Because we have control over which kinds of form we use in each step in our wizard, this lets us:

- be pretty explicit about the overriden behaviour in code
- write tests fairly to check for this modified behaviour
- limit where the modified behaviour is taking place, so we dont need to keep checking for it elsewhere

The main thing I'm not sure about how file uploads are handled

mrchrisadams:
I've implemented this change in #527.

Once I've merged that into this PR, assuming all the tests pass, I'll merge this in, as I think this will give us something we can test in staging.

PR #524, opened 2023-10-06
IP range text change

hanopcan:
Text change request from @fershad 

More details here: https://trello.com/c/JZVQsTSb/256-change-help-text-on-ip-range-part-of-verification-form


hanopcan:
@mrchrisadams This is a simple copy change, but I'd still appreciate a quick QA if poss. Thanks :smile:

PR #527, opened 2023-10-18
Add support for updating hosting providers when not all forms change in wizard

mrchrisadams:
This PR picks up the earlier work in #522, to address the problems we saw when updating an existing provider using the form wizard.

The problem is detailed in that PR, but the short version is that some data in forms contained within django formsets were not making it through the form wizard, which means when updating a provider request, if there were forms in a formset that had no changes, they were being dropped.

This was a blocker as it meant that on approval, our provider would be updated with incomplete information.

This only seemed to happen when our wizard form contained formsets that contained forms with unchanged data, and after exploring various options, I arrived at the idea of introducing a mixin class for use in forms we use in the form wizard.

All it does is adapt form behaviour to always return `True` when a `has_changed()` method is called on a form.

Because the formset handling code relies on this check when deciding to include a form in the validation steps for a given formset, this will always include the form in the data that is saved by the formset.

When used in a multi stage wizard, this ensures that the all the form wizard stages that contain formsets  also retain information in the forms even if no change has taken place. When the verification request is approved, all the data is saved back to the hosting provider.

### DIFF
--- a/apps/accounts/forms.py
+++ b/apps/accounts/forms.py
@@ -25,6 +25,21 @@ from django.utils.safestring import mark_safe
 
 User = get_user_model()
 
+class AlwaysChangedModelFormMixin:
+    """
+    A mixin for ModelForms that makes sure that a form with this Mixin
+    is always marked as changed in checks like `has_changed()`.
+    Used in form wizards containing formsets, to return true for all forms 
+    in a formset, so that all data is saved at the end of the wizard.
+    """
+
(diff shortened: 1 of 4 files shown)


mrchrisadams:
[Users should be able to update their existing hosting provider listings](https://trello.com/c/l5g3ghlA/286-users-should-be-able-to-update-their-existing-hosting-provider-listings)

PR #529, opened 2023-10-20
Hs editing provider front end

hanopcan:
Some front-end changes as we implement the verification and provider updating processes.

This relates to this Trello card: https://trello.com/c/VdEJxyna/291-show-verification-requests-at-the-top-when-more-information-is-needed-and-tidy-up-homepage-generally


hanopcan:
@mrchrisadams I'm assigning this to you to take a look at and merge in.

Because the new hosting provider functionality is hidden behind a flag, I don't believe this needs to wait. Deploying this now should deliver some improvements to the existing portal home, which will be helpful for Fish and users. And then when we deploy the provider editing stuff and lift the flag, the new edit listing button will become visible too.

So basically as long as I haven't done anything wild in the code I think we should merge now.

apps/accounts/templates/base.html
```diff
@@ -81,6 +81,7 @@ <h1 class="text-5xl">Django + Tailwind = ❤️</h1>
 			<div class="container mx-auto px-2 sm:px-4 md:p-0 prose">
 				<div class="p-6 bg-white border-2 rounded-xl">
 					<h3 class="text-xl uppercase">Need help?</h3>
+					<p>Common questions and their answers - <a href="https://www.thegreenwebfoundation.org/support/" target="_blank" rel="noopener noreferrer">browse our FAQs</a>.</p>
```
mrchrisadams, line 84: Can you add a little about what the basis is for having this open in a new tab? When I worked with Ross, he started doing this neat thing in PRs where he would add comments to the lines in the PR to explain the reasoning for a change when it wasn't obvious.

I think I can understand the basis for this in that it's supplementary if this was in the workflow on a form wizard, but this looks like it would turn it lots of places.

We don't have any formal guidance for this right now, and it's often pretty subjective, so having a consistent approach we can refer to would be useful.

I'm thinking of the guidance below for example
https://www.nngroup.com/articles/new-browser-windows-and-tabs/

https://www.w3.org/WAI/WCAG21/Techniques/general/G200

Is this kind of stuff what you were thinking? In that it's ancillary to the task at hand of submitting a verification request?

hanopcan: Thanks Chris! Yes in this instance that was my logic. The link is taking you away from the provider portal and to an entirely separate site, which as you say could lead to all sorts of places/rabbit holes. My general understanding of opening things in new tabs is its good practice if the user is being taken to another domain entirely. In this case they are. I don't have strong feelings about this, and if this is the only blocker I'm happy to defer to whatever you feel is best :smile:

PR #542, opened 2024-01-05
Update green badges

hanopcan:
https://trello.com/c/XAMBEfu6/324-rebrand-green-web-check-badges-technically


mrchrisadams:
As discussed in trello, I've changed this PR bit to:

1. keep support for the old image urls, so anyone hyperlinking to old images doesn't get a surprise
2. serve the updated images at a `v3` endpoint like the other API endpoints
3. refactor the code so the image generation is handled in its own file, and the views just call methods on the corresponding image generators

PR #546, opened 2024-01-20
WIP: Add Marimo as a dependency

mrchrisadams:
This is a PR to merge in Marimo, a software library that makes carrying out anaylsis of greenweb data using the ORM made available by Django.

We've use it like so in development, when working with a working database connection:

```
pipenv run marimo edit ./scripts/path-to-analysis-file.py
```

The scripts would be kept in a file that is listed in `.gitignore`, or potentially a git submodule.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -21,7 +21,6 @@ hypothesis = "*"
 faker = "*"
 black = "*"
 factory_boy = "*"
-rich = "*"
 sphinx = "*"
 sphinx-autobuild = "*"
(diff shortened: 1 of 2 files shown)


mrchrisadams:
This works with a separate git repo under `data-analysis` fairly nicely, and having the the separation is already useful. I'm using this like this happily:

```
pipenv run marimo edit ./scripts/path-to-analysis-file.py
```

I'm deffo think this is worth merging in.

PR #550, opened 2024-02-12
Add sentry to portal frontend

fershad:
This PR adds some Sentry error monitoring code to the frontend of the platform so that we can better placed to understand and investigate frontend errors being reported by users.

Uses code from:
https://docs.sentry.io/platforms/javascript/
https://docs.sentry.io/platforms/javascript/troubleshooting/#capturing-resource-404s

Currently, the main purpose of this implementation is to capture information about a long running, sporadic CSS loading issue that has been experienced on the frontend.

### DIFF
--- a/apps/accounts/templates/base.html
+++ b/apps/accounts/templates/base.html
@@ -9,57 +9,59 @@
         <meta http-equiv="X-UA-Compatible" content="ie=edge">
         {% comment %} Load sentry before any CSS/other JS loads {% endcomment %}
 
-        <script
-            src="https://js.sentry-cdn.com/dd30eb65f140420688f813ab96721a1c.min.js"
-            crossorigin="anonymous"
-        ></script>
+        {% if sentry %}
+            <script
+                src="https://js.sentry-cdn.com/dd30eb65f140420688f813ab96721a1c.min.js"
+                crossorigin="anonymous"
+            ></script>
 
-        <script>
-            window.sentryOnLoad = function () {
-                Sentry.init(
-                    dsn: "{{ sentry.dsn }}",
-                    //   This should be updated with each release.
-                    // TODO: Is there a way to automate this?
-                    release: "{{ sentry.release }}",
-                    integrations: [
-                        Sentry.browserTracingIntegration(),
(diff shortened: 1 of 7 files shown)


apps/accounts/templates/base.html
```diff
@@ -7,6 +7,62 @@
 		<meta charset="UTF-8">
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta http-equiv="X-UA-Compatible" content="ie=edge">
+		{%- comment -%} Load sentry before any CSS/other JS loads {%- endcomment -%}
```
mrchrisadams, line 10: Would you change the  `{%- comment -%}` to be `{% comment %}` ?

I think this `{%-` formulation might be used templating languages in nunjucks, but unfortunately it's not supported in django's own templating language

apps/accounts/templates/base.html
```diff
@@ -7,6 +7,62 @@
...
+
+<script
+  src="https://js.sentry-cdn.com/dd30eb65f140420688f813ab96721a1c.min.js"
+  crossorigin="anonymous"
+></script>
+
+<script>
+  window.sentryOnLoad = function () {
+    Sentry.init({
+      dsn: "https://dd30eb65f140420688f813ab96721a1c@o14742.ingest.sentry.io/2071988",
```
mrchrisadams, line 21: I think this would result in every environment (staging/ prod etc) we have reporting to the same dashboard in sentry. I'd prefer this was variable passed in if possible, but I appreciate it's a thing I'm probably better placed to add, as we'd like rely on this being an environment variable or similar.

mrchrisadams (commented):
Hi @fershad  - thanks for this! I think I understand the changes here and the feedback I have are mainly niggles or things I'm best off changing at my end.

I'm happy to merge this in, and I've highlighted the small changes I might make.

apps/accounts/templates/base.html
fershad: Cheers Chris. Guidance here in the Sentry docs for that:
https://docs.sentry.io/platforms/javascript/configuration/environments/

PR #567, opened 2024-03-12
Revise gitpod

mrchrisadams:
This PR updates the Gitpod file to remove the unused Redis process, and update the rabbitMQ image used in development.

Previously, RabbitMQ was failing with the following error:

```
Failed to create thread: Operation not permitted (1)
```


mrchrisadams:
Hmm.. the rabbitMQ still needs some love. 

I initially looked into this here:
https://stackoverflow.com/questions/76703987/rabbitmq-docker-failed-to-create-thread-operation-not-permitted-1

I suspect we'll need some way to run this as  `--privileged`

PR #575, opened 2024-04-24
Fix ipv6 support in domain lookups

### LINKED ISSUE #573: convert_domain_to_ip does not support ipv6, only ipv4
We curently have a method in our platform that is responsible for converting domain names to ip address objects in python.

It's linked below:
https://github.com/thegreenwebfoundation/admin-portal/blob/master/apps/greencheck/domain_check.py#L128

The problem is that right now, it only supports ipv4, when it should support ipv6 too. the code looks like this:

```python


    def convert_domain_to_ip(
        self, domain
    ) -> ipaddress.IPv4Network or ipaddress.IPv6Network or None:
        """
        Accepts a domain name or IP address, and returns an IPV4 or IPV6
        address
        """
        try:
            ip_string = socket.gethostbyname(domain)
            return ipaddress.ip_address(ip_string)
        except socket.gaierror as err:
            logger.warning(f"Unable to lookup domain: {domain} - error: {err}")

```

This also method right now unfortunately  will return a `None` value when `socket.gethostbyname()`, which causes a problem later in the code path. 

Really we should be using `socket.getaddrinfo`, which _does_ work with ipv6, but instead of just returning a string representation of an ip address, returns a different object., which we'd need to change. the docs for python are below:

https://docs.python.org/3.12/library/socket.html#socket.getaddrinfo

We'd need `convert_domain_to_ip` to return a string representation of ipv4 addresses AND ipv6 addresses, which involves a bit of work to fetch the string out of the return value of `socket.getaddrinfo`.

When there is no IPv4 or IPv6 returned, we should probably raise an appropriate exception, so we can catch it early, rather than blindly passing `None` values along to the caller of the method.

### What the code returns

I've put together a quick marimo notebook example of what the `socket.getaddrinfo` method returns, to give an idea of what datastructures any code would need to fetch the string representation of an IP address out of:

https://static.marimo.app/static/get-addrinfo-check-e861

This isn't mega complex, but it's likely a few more lines than we have in the file - the good news it doesn't rely require existing knowledge of the rest of the app, so presumably anyone with a bit of python experience could make a PR with an improved implementation of `convert_domain_to_ip`.

mrchrisadams:
This PR is intended to resolve #573 , which came up when we realised that domain lookups that resolve to an ipv6 address were not working properly.

It covers both cases, and lays the path for a updating this to check ALL ip addresses a domain might resolve to in future, not just the first one. 

We don't immediately moved to checking every domain, as I'm not sure of the performance implications, and there are a few places that expect just one ip address to be returned - we'd need to make quite a few changes elsewhere.

### DIFF
--- a/apps/greencheck/domain_check.py
+++ b/apps/greencheck/domain_check.py
@@ -16,8 +16,10 @@ This follows largely the same approach:
 import ipaddress
 import logging
 import socket
+import typing
 import urllib
 
+import ipwhois
 import tld
 from django.utils import timezone
 from ipwhois.asn import IPASN
@@ -28,10 +30,6 @@ from .choices import GreenlistChoice
 from .models import GreenDomain, SiteCheck
 
 logger = logging.getLogger(__name__)
-
-import ipwhois
-
-import typing
 
 
 class GreenDomainChecker:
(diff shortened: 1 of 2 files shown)


apps/greencheck/domain_check.py
```diff
@@ -228,13 +256,21 @@ def check_domain(self, domain: str) -> SiteCheck:
         the best matching IP range for the ip address it resolves to,
         or a 'grey' Sitecheck
         """
+        UNRESOLVED_ADDRESS = "0.0.0.0"
```
mrchrisadams, line 259: I'm aware that 0.0.0.0 has a special meaning when we pass it to servers. I'm using this placeholder, as I'm not aware of an equivalent IP address that means "not a real address", and in a few places we always assume an ip address string is returned.

apps/greencheck/domain_check.py
```diff
@@ -127,16 +129,42 @@ def asn_from_ip(self, ip_address):
...
-            logger.warning(f"Unable to lookup domain: {domain} - error: {err}")
+
+        # TODO: support multiple addresses being returned:
+        # `getaddrinfo`` actually returns a list of possible addresses,
+        # but our current code only assumes a domain would resolve to a single IP
+        # address when we look up a domain.
+        # Ideally we'd check that ALL the IP addresses resolved are within our
+        # green IP ranges but until we know how much this impacts performance
+        # we choose the first one.
+        ip_info = socket.getaddrinfo(domain, None)
```
mrchrisadams, line 145: As the comments around this say - we acknowledge that we really should check that every ip range is green. Ths priority is fixing this first, before adding that functionality in later.

mrchrisadams:
[Fix IPv6 support for greencheck](https://trello.com/c/PoGzOnhR/350-fix-ipv6-support-for-greencheck)

mrchrisadams:
[Fix IPv6 support for greencheck](https://trello.com/c/PoGzOnhR/350-fix-ipv6-support-for-greencheck)

PR #578, opened 2024-04-25
Update CSV importer usage in admin form

mrchrisadams:
While we updated the CSV importer, we didn't update how it was used in the admin.

This changes the form code to create an instance of the new style importer, that share the revised structure with the AWS, Google and Azure importers.


mrchrisadams:
[Fix CSV Importer in django admin](https://trello.com/c/wjyqOYK2/349-fix-csv-importer-in-django-admin)

PR #581, opened 2024-05-02
Update approve flow handle updated evidence as well as new evidence

mrchrisadams:


apps/accounts/models/provider_request.py
```diff
@@ -282,26 +286,51 @@ def approve(self) -> Hostingprovider:
                 hostingprovider=hp,
             )
 
+        # Fetch our archived documents:
```
mrchrisadams, line 289: This represents a change in our logic.

Because submitted evidence in a verification request and supporting evidence are different kinds of objects, with different IDs, even if they contain the same content, we need a way to see they are "functionally equivalent", to determine if they should be retained on the hosting provider object after an approval.

apps/accounts/models/provider_request.py
```diff
@@ -282,26 +286,51 @@ def approve(self) -> Hostingprovider:
...
         # create related objects: supporting documents
         for evidence in self.providerrequestevidence_set.all():
+            logger.debug(f"checking for matching evidence for: {evidence}")
+
             # AbstractSupportingDocument does not accept null values for `url`
             # and `attachment` fields
             url = evidence.link or ""
             attachment = evidence.file or ""
 
-            # check for the existence of archived evidence. Unarchive if found, like we do
```
mrchrisadams, line 292: This is the key change - instead of running this check for every iteration of the loop, we fetch all the archived supporting documents once, and check the evidence against them, where the logic us defined in `is_already_uploaded` rather than the ORM query.

We do this because we don't have a reliable, fast way yet to compare the contents of the files from an ORM query.

apps/accounts/models/provider_request.py
```diff
@@ -403,3 +446,29 @@ def clean(self) -> None:
             raise ValidationError(f"{reason}, you haven't submitted either.")
         if self.link and bool(self.file):
             raise ValidationError(f"{reason}, you've attempted to submit both.")
+
+    def has_content_match(self, other_doc: "HostingProviderSupportingDocument") -> bool:
```
mrchrisadams, line 450: This is where we push much of the logic for comparing submitted evidence to already uploaded supporting documents.
This currently only supports checking evidence against supporting docs - not the other way around, and not evidence against evidence, or supporting doc against other supporting docs.

apps/greencheck/tests/views/test_stats_view.py
```diff
@@ -207,6 +210,7 @@ def test_stat_view_top_domains(
 
         assert len(top_green_domains) == 10
 
+    @pytest.mark.flaky
```
mrchrisadams, line 213: This code isn't really used, and we don't expose the stats anywhere. I've commented out the tests for now, and it might make sense to remove it entirely in a separate PR if we're not exposing this to end users.

apps/accounts/models/provider_request.py
```diff
@@ -227,6 +231,9 @@ def approve(self) -> Hostingprovider:
         hp.request = self
         hp.created_by = self.created_by
 
+        # if we have approved a submission from a provider
```
mrchrisadams, line 234: We are unarchiving all the related items, but until now, we were still leaving the provider archived. Approving the provider is an explicit acceptance of their data, so we also unarchive the provider itself now.

mrchrisadams:
[When a provider's verification request is approved, make sure the provider is unarchived.](https://trello.com/c/Z7kYc292/359-when-a-providers-verification-request-is-approved-make-sure-the-provider-is-unarchived)

PR #582, opened 2024-05-03
Audit logs for creation and updates to verification requests

mrchrisadams:
This PR introduces use of the django log entry feature to provide a more accessible audit log when staff or users are working with verification requests.

### DIFF
--- a/apps/accounts/admin.py
+++ b/apps/accounts/admin.py
@@ -7,7 +7,9 @@ from dal_select2 import views as dal_select2_views
 from django import template as dj_template
 from django.conf import settings
 from django.contrib import admin, messages
+from django.contrib.admin.models import CHANGE
 from django.contrib.auth.admin import Group, GroupAdmin, UserAdmin
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import PermissionDenied
 from django.core.mail import send_mail
 from django.db import models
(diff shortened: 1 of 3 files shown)


apps/accounts/admin.py
```diff
@@ -1448,6 +1450,25 @@ def send_approval_email(
             template_html="emails/verification_request_approved.html",
         )
 
+    def log_message(
```
mrchrisadams, line 1453: This convenience method is intended to make it much easier to log changes we make that might not otherwise be made in the admin

apps/accounts/views.py
```diff
@@ -279,6 +283,26 @@ class Steps(Enum):
         Steps.PREVIEW.value: "provider_registration/preview.html",
     }
 
+    def log_creation(
```
mrchrisadams, line 286: This allows us to log various events to the django log_entries table, from _outside_ the admin.

It's used outside approve() method for a provider request because access to the user in a given request is required.

If we created a "robot" user, we coud use this same approach for logging automated updates too, but that's outside the scope of this PR.

mrchrisadams:
[Updated evidence in provider requests not appearing in hosting provider listing after approval](https://trello.com/c/NNnRai9W/355-updated-evidence-in-provider-requests-not-appearing-in-hosting-provider-listing-after-approval)

PR #590, opened 2024-05-18
Eco-CI integration

ArneTR:
Hey @mrchrisadams ,

here is the promised Eco-CI integration sample PR!

What I did:
- Integrate Eco-CI initially and initalized it also with CarbonDB. Your runs will then not only show up here in the Eco-CI view, but also in Carbon DB
  - Eco-CI View: 
  - CarbonDB view: https://metrics.green-coding.io/carbondb-lists.html?company_uuid=7f1d34f0-7666-4378-8d8f-d37cced7ccb0
I have assinged the company UUID `7f1d34f0-7666-4378-8d8f-d37cced7ccb0` for your company. Note that you can use any ID. You should just double check that nobody else is using it.

- I integrated a measurement step after every step in the workflow. This way you get a nice granular view of all steps and how they relate in consumption

- I integrated `send-data: true`. This will send data here: https://metrics.green-coding.berlin/ci-index.html
  - In case you do not want that set it to false

- I also activated PR view, which makes a nice comment in the PR. This however needs the permission to write to the PR. If you do not want that please remove it.
  - We typically allow it but set our PRs to only run after they have been approved by a member of the organization. I can recommend that in general if you do not have it active

- Also I added an idea how to reduce unnecessary runs. I added some file exclusions. It will now not run when any changes to Markdown files occur, which I assume are non-functional in your repo?
  - If you do not want it remove the line. But maybe you want to add more even :)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,10 @@
 name: Run tests
 
-on: [push]
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+      - '.gitignore'
 
 defaults:
   run:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -34,16 +34,43 @@ jobs:
         python-version: [3.11]
 
     steps:
+      - name: Eco CI Energy Estimation - Initialize
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: start-measurement
+          company-uuid: "7f1d34f0-7666-4378-8d8f-d37cced7ccb0"
+          project-uuid: "a9946d61-149b-4757-99e5-aaac24acf289"
+        continue-on-error: true
+
       - uses: actions/checkout@v4
+      - name: Eco CI Energy Estimation - Get Measurement
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: get-measurement
+          label: 'checkout'
+        continue-on-error: true
 
       - name: Use Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Eco CI Energy Estimation - Get Measurement
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: get-measurement
+          label: 'setup-python'
+        continue-on-error: true
+
       - name: Install tooling for managing dependencies
         run: |
           python -m pip install --upgrade uv wheel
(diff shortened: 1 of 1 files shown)


mrchrisadams:
Hey @ArneTR , this is so cool! I haven't forgot about it - I've just had my head down figuring out how to use GMT. I'll add some feedback either this afternoon or tomorrow.

.github/workflows/run-tests.yml
```diff
@@ -1,6 +1,10 @@
 name: Run tests
 
-on: [push]
+on:
+  push:
+    paths-ignore:
+      - '**.md'
```
mrchrisadams, line 6: This, and the line below sets CI to avoid running if we're making changes to documentation, or in the line below the `.gitignore` file. We almost definitely to not want to trigger a deploy and CI run in these cases so it makes sense to have them.

.github/workflows/run-tests.yml
```diff
@@ -34,16 +40,43 @@ jobs:
...
+      - name: Eco CI Energy Estimation - Initialize
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: start-measurement
+          company-uuid: "7f1d34f0-7666-4378-8d8f-d37cced7ccb0"
+          project-uuid: "a9946d61-149b-4757-99e5-aaac24acf289"
+        continue-on-error: true
+
       - uses: actions/checkout@v4
+      - name: Eco CI Energy Estimation - Get Measurement
```
mrchrisadams, line 52: This step extra step is measuring the checkout stage. If we wanted to measure other steps in the build process we'd add the `green-coding-solutions/eco-ci-energy-estimation@main`, but presumably once we have the project and org id set, we needn't repeat it in later steps

.github/workflows/run-tests.yml
```diff
@@ -80,3 +119,17 @@ jobs:
...
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: get-measurement
+          label: 'pytest'
+        continue-on-error: true
+
+      - name: Eco CI Energy Estimation - End Measurement
+        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        with:
+          task: display-results
```
mrchrisadams, line 132: this is the other predefined task  that the `green-coding-solutions/eco-ci-energy-estimation@main` seems to offer. It's what creates out neat charts

mrchrisadams (approved):
Thanks for this @ArneTR - I think I have a good understanding of this code is doing, I'm merging this in.

I also learned a few neat tricks about GH actions I didn't know before - danke!

PR #593, opened 2024-05-28
Fix the django autocomplete / select2 upgrade bug

mrchrisadams:
This PR introduces two things:

1. an override of the hosting provider change_form, to allow us direct control over which scripts are loaded by the page, and in what order
2. a concatenated javascript file, created to overcome the incredibly frustrating clash of different versions of jQuery, select2 and other scripts being loaded onto the page, that led to the autocomplete tag widget failing to load properly since we upgraded Django.

There is no magic for the concatenated file, and I am not using any bundling tools - it's literally a file where I have pasted the files in the order we need them to load, and will need to be updated when we make any further changes to this hosting provider admin page.

We're in the process of removing access to these django admin pages to non-staff users - so while the new 'compiled' javascript file makes for a large uncompressed file, the difficulty of fixing this bug has led me to err towards having a slower loading, but easier to debug page for when we have to come back to this.

### DIFF
--- a/apps/accounts/static/accounts/js/compiled.select2.js
+++ b/apps/accounts/static/accounts/js/compiled.select2.js
@@ -1,0 +1,18581 @@
+/**
+ * This is concatenated set of the libraries on the admin hosting provider change form
+ * that when loaded in a production environment, fail to load in the correct order
+ * resulting in errors because jQuery or the `$` variable is undefined at the time of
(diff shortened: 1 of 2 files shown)


mrchrisadams:
[Fix provider labels in backend](https://trello.com/c/pzhfXgUE/362-fix-provider-labels-in-backend)

PR #598, opened 2024-06-14
Add extra docs after pairing on set up with Han

mrchrisadams:
This PR adds to the docs for people coming to the codebase for the first time, and adds extra information to the makefile


mrchrisadams:
@hanopcan  I've made the updates we discussed on the call. Would you mind opening a PR and merging in any other changes if there was stuff that ought to be clearer?

There's no need to wait for review from me to merge it in, as these would mainly be changes for our future selves or future team members.

PR #602, opened 2024-06-18
Moving to v4.0-rc3 and added machine-uuid

ArneTR:
Hey @mrchrisadams ,

we have massively slimmed our Eco-CI tool and wanted to bring you the lastest update via PR.

## What I changed in your workflow
- Moving from `@main` inclusion to `@VERSION`. This will instruct Dependabot to also make PRs automatically in the future in case we update Eco-CI.
- Using the newest super slim v4.0-rc3 version. Find all details like a 100x improvement here: https://github.com/green-coding-solutions/eco-ci-energy-estimation/releases/tag/v4.0-rc3
- We now require `machine-uuid` to be set. I generated a random one for you that you can then find in your CarbonDB view: https://metrics.green-coding.io/carbondb-lists.html?company_uuid=7f1d34f0-7666-4378-8d8f-d37cced7ccb0
  - Note there: They way this was defined in the PR was sadly incorrect. Probably my bad. But Eco-CI did not pick up the CarbonDB integration. It should now!

Also we have done some improvements to the SCI calculation and are now showing more granular data.

In essence though the speed gain should be the biggest win. We went down from ~30 seconds to ~1-2 seconds of overhead for a typical 5-7 step workflow.

## PR-Comments

Somehow the PR comments are not showing correctly for you. I am not sure why that is. For every PR you should get a comment like that: 
Screenshot 2024-06-18 at 2 49 59 PM

I have a feeling some general setting on the repo is more restrictive than we expect it to be. I will look into that once a workflow for this PR has run

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -41,16 +41,17 @@ jobs:
 
     steps:
       - name: Eco CI Energy Estimation - Initialize
-        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4.0-rc3
         with:
           task: start-measurement
           company-uuid: "7f1d34f0-7666-4378-8d8f-d37cced7ccb0"
           project-uuid: "a9946d61-149b-4757-99e5-aaac24acf289"
+          machine-uuid: "90a9ec92-0884-43bc-8b98-9a7885864307"
         continue-on-error: true
 
       - uses: actions/checkout@v4
       - name: Eco CI Energy Estimation - Get Measurement
-        uses: green-coding-solutions/eco-ci-energy-estimation@main
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4.0-rc3
         with:
           task: get-measurement
           label: 'checkout'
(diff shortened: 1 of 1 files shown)


ArneTR:
Actually by submitting this PR I might have found the issue:

- Your https://github.com/thegreenwebfoundation/admin-portal/blob/master/.github/workflows/run-tests.yml file only specifies to run on `push`. So me making a PR does not trigger that workflow as I am an outside collaborator and technically never *pushed*. 

Is that what you want?

In case you want that I can add it to the PR. Let me know. In any case it would be needed if you want the Eco-CI PR-Comment funcationality as I believe there is no safe way to determine if we are in a PR event if you have not set the trigger in the workflow.

mrchrisadams:
Hi @ArneTR 

> Your https://github.com/thegreenwebfoundation/admin-portal/blob/master/.github/workflows/run-tests.yml file only specifies to run on push. So me making a PR does not trigger that workflow as I am an outside collaborator and technically never pushed.

I think this is _might_ be related to the act that we've been pushing branches _inside_ our own repo when we do PRs, rather than working with PRs from external repos.

TBH, I think it probably does make sense to run this on external PRs too as you suggest. I've never done this before though - would it just be a case of adding something like this?

```yaml
on:
  push:
    paths-ignore:
      - '**.md'
      - '.gitignore'
  pull_request:
    paths-ignore:
      - '**.md'
      - '.gitignore'
```

I'm not sure what the correct syntax would be. If you know, would you add the correct incantation? I'll merge it in once it's made as I think this would be pretty neat.

ArneTR:
jap, that looks correct

PR #603, opened 2024-06-19
Remove unused github actions

mrchrisadams:
This PR removes github actions that we disabled before because they would either timeout, or runout or memory.

We've since switch to running these internally.


ArneTR:
yay, pr display works! ❤️

Also the carbonDB data is showing https://metrics.green-coding.io/carbondb-lists.html?company_uuid=7f1d34f0-7666-4378-8d8f-d37cced7ccb0

mrchrisadams:
Nice, thanks for the PR @ArneTR.

PR #616, opened 2024-08-23
Improve directory filtering to exclude archived providers

fershad:
This tiny PR improves the filtering of providers when the Directory list is rendered.

The previous logic would only exclude providers who were marked as `showOnWebsite=false`. This exposed an edge case where providers might be archived (`archived=true`) however might still have the showOnWebsite variable set to true (`showOnWebsite=true`). In this case the provider would incorrectly be shown on the Directory even though they are archived.

This PR fixes that.

### DIFF
--- a/apps/greencheck/views.py
+++ b/apps/greencheck/views.py
@@ -2,7 +2,6 @@ import collections
 import typing
 from datetime import date, timedelta
 
-import waffle
 from dateutil.relativedelta import relativedelta
 from django.db.models import Sum
 from django.http import HttpResponseRedirect
@@ -14,7 +13,6 @@ from django.views.decorators.cache import cache_page
 from django.views.generic.base import TemplateView
 
 import django_filters
-from waffle.mixins import WaffleFlagMixin
 
 from apps.accounts.models.hosting import Hostingprovider
 from apps.greencheck.models.checks import GreenDomain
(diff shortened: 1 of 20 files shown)


apps/greencheck/views.py
```diff
@@ -235,21 +235,23 @@ class Meta:
...
         """
         Populate the page context with the filtered list
         of providers
         """
 
-        queryset = Hostingprovider.objects.filter(showonwebsite=True).prefetch_related(
+        queryset = Hostingprovider.objects.filter(
+            showonwebsite=True,
+            archived=False
+        ).prefetch_related(
```
fershad, line 254: @mrchrisadams The only change I made was here. But this PR is showing a bunch of other files being changed too. Is that because I'm merging it into Staging rather than Main?

mrchrisadams: hi @fershad  - I think it might be because what's on staging might not be the same as main / master.

I've typically treated staging as something we're ok to `git push --force` rather than open up PRs to `staging`. 

My reasoning is that while we care about a specific branch being merged into `master`, there are sometimes cases where we might want to see how something looks in `staging`, without then proceeding directly into  merge it into `master`, and we also make no guarantees about `staging` always working. 


```
git push origin ca-card-350-fix-ipv6-support:staging -f                                                                                                                                           
```

and in this case we'd do something like

1. update this PR to be a PR against `master`

2. force push the change to staging with something like the command below

```
git push origin improve-directory-filter:staging -f                                                                                              
```

I'll try to respond in Zulip today if you DM me, otherwise we have the call on Monday to chat as well.

mrchrisadams: @fershad  are you ok with updating the PR so it's a change against master, a bit like the way you have a comparison view here?

https://github.com/thegreenwebfoundation/admin-portal/compare/master...add-disclaimer-for-dataset

PR #625, opened 2024-11-13
Update github codespace and justfile

mrchrisadams:
This PR updates the devcontainer file used to define what is built when spinning up a codehub codespace.

It also adds a slightly more convenient Justfile, for running django management commands and using Marimo for data analysis


mrchrisadams:
ZMG. It looks like adding this triggers the proper build steps!

```json
"customizations": {
    "codespaces": {},
}
```

PR #630, opened 2024-12-24
Multiple small quality of life fixes

fershad:
This PR: 

- Adds a max-length of 1000 characters to the Provider Description textarea field in the verification request form.
- Adds an overflow property to long file names in file input fields.
- Restricts the types of files a user can submit in a file input to only images & PDFs.
- Update terminology across the repository.
- Update wording on initial verification page for consistency.

### DIFF
--- a/apps/accounts/templates/emails/migration_email.txt
+++ b/apps/accounts/templates/emails/migration_email.txt
@@ -1,4 +1,4 @@
-Dear member of The Green Web Foundation,
+Dear member of Green Web Foundation,
 
 Thank you for the contributions that you've made in the past to our hosting database. We are contacting you now because we need you to do something to be able to access the database in the future. After eight years of good service, the administration system behind our database needed an upgrade. The good news is that the upgrade succeeded. However, due to a new password hashing algorithm, we are obliged to ask all members to reset their password.
 
@@ -8,4 +8,4 @@ Feel free to contact us at support@thegreenwebfoundation.org at any time if you 
 
 Thanks in advance for your help!
 With best regards,
-The Green Web Foundation team
+Green Web Foundation team
--- a/apps/accounts/templates/emails/migration_email.txt
+++ b/apps/accounts/templates/emails/migration_email.txt
@@ -1,4 +1,4 @@
-Dear member of Green Web Foundation,
+Dear member of the Green Web Foundation,
 
 Thank you for the contributions that you've made in the past to our hosting database. We are contacting you now because we need you to do something to be able to access the database in the future. After eight years of good service, the administration system behind our database needed an upgrade. The good news is that the upgrade succeeded. However, due to a new password hashing algorithm, we are obliged to ask all members to reset their password.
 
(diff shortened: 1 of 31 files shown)


fershad:
@mrchrisadams This felt like a pretty straightforward change to make, because there's a similar `max_length` setting applied to the organisation name field in the form (on the same page as the description).

However, when testing with a reeeaaaaallllly long organisation name, I found that the form will just truncate the name to 255 characters without giving any warning. I'm okay with that functionality, but want to make a note that is what will happen with very long descriptions too.

apps/accounts/templates/emails/migration_email.txt
```diff
@@ -1,4 +1,4 @@
-Dear member of The Green Web Foundation,
+Dear member of Green Web Foundation,
```
mrchrisadams, line 1: would you mind using the lower case the here instead removing it @fershad ? I think it reads funny, as there's only one Green Web Foundation. I've expanded on this here, and referenced it in the controlled vocabulary:

https://docs.google.com/document/d/1ag3WJ01saMP0NO1-WHD0xq0DpC9koTzVnH3I8OUUWU8/edit?tab=t.0#heading=h.ctxl9icrf6iv

apps/greencheck/swagger.py
```diff
@@ -18,7 +18,7 @@
 
 schema_view = get_schema_view(
     openapi.Info(
-        title="Welcome to The Green Web Foundation Partner API ",
+        title="Welcome to Green Web Foundation Partner API ",
```
mrchrisadams, line 21: Same here as above. Can we please keep the lower case 'the' here, like we do in other places where we refer to the Green Web Foundation in the middle of a sentence?

docs/installation.md
```diff
@@ -1,16 +1,19 @@
 # Installation
 
-This installation shows you how to setup the Admin Portal of The Green Web Foundation.
```
mrchrisadams, line 3: As above.

docs/key-concepts.md
```diff
@@ -1,6 +1,6 @@
 # Key concepts
 
-The Green Web Foundation offers an API to provide any IP address or domain, and get back a 'green' result indicating that the site or digital service appears to be running on digital powered by green energy, and we have found evidence to support this claim or a 'grey' result where no evidence was found for green energy.
```
mrchrisadams, line 3: I think it's ok to have the 'The' here to be honest, as we're talking about the foundation offering an API, and it's sentence in a paragraph, rather than a the name of the organistion all by itself. Personally, I think it reads a bit funny when we remove the article. This is consistence with the usage I've outlined in the article, where the Mozilla Foundation, another foundation is totally with beginning sentences with 'The' if it makes them read better. I've linked directly to the example below:

https://docs.google.com/document/d/1ag3WJ01saMP0NO1-WHD0xq0DpC9koTzVnH3I8OUUWU8/edit?tab=t.0#heading=h.t4lcbe6foxja

mrchrisadams (approved):
Thanks @fershad  - I've added a few comments but otherwise this look good, thanks!

PR #640, opened 2025-03-07
Add compose and devcontainer for running app in github codespaces

mrchrisadams:
This PR introduces changes to allow us to migrate from gitpod  and their gitpod.yml file, to the newer devcontainer.json approach used by github codespaces, and other tools like the newer gitpod flex, and devpod.

It introduces a minimal docker compose file to run the backing services, and an updated devcontainer file with various helpful plugins for vs code, and providing toolsl like `uv` and `just`

To make this work with github spaces, we'd need 

the following

- [x] figure out how to run multiple containers like we do with gitpod.yml (one dev container, one rabbitmq, one mariadb container)
- [x] migrate set up command for main dev container to run migrations
- [x] update the docs to refer to devcontainer and probably github codespaces
- [x] fix tests - the failing tests seem to be related to carbon.txt changes we can get away with separating out from this work. If we haven't fixed them with a different PR, then we might be able to just skip them for now, as the implemntation of carbon.txt parsing in this app is likely out of date now that we have a standalone validation library

### Tradeoffs this includes

At present, we have a `just` task to support the use of Marimo for local data analysis in a notebook style environment. This relies on being able to access another private repo, https://github.com/thegreenwebfoundation/data-analysis. Github Codespace's security model makes checking out other github repos a bit of a pain, so using a Marimo notebook will need to happen locally for now. The main poeple who use these directly are myself and @fershad, and there are workarounds which are doable, but would make this PR huge. I've created #644 as a bookmark for this idea.

### extra stuff, might be out of scope

This essentially replaces Gitpod one hosted dev environment provider, for codespaces, another hosted dev environment provider. Ideally we would _not_ be tied to any one provider, as `devcontainer.json` is supposed to be a standard.

One option is to target [devpod](https://devpod.sh/), which appears to be the open source equivalent to gitpod/ codespaces, and from what I can tell, appears to be something like a default you could run on any infra. It would _also_ allow for running a setup locally if you have something like docker installed, so you wouldn't need to be connected to the internet for development.

However, for this specific app, the development container we rely on, the `mcr.microsoft.com/devcontainers/universal:2` container only works with `x86` architecture. If you have a recent mac running Apple Silicon, then this dev container will not work with `arm64` based processors for local development.

You can see more about the container, and which systems it targets below

https://github.com/devcontainers/images/blob/main/src/universal/history/2.9.0.md

#### Finding a more compatible container

One option might be to use a different dev container that _does_ target `arm64` architecture. However, if we do this, we'd still need to figure out what extra things the `universal` container is relying on to work with multiple containers and `docker-compose` files.

https://github.com/devcontainers/images/blob/main/src/python/history/1.2.1.md

----

### Auto generated summary

This pull request includes significant changes to transition the development environment from Gitpod to GitHub Codespaces. The primary changes involve updating configuration files, environment variables, and documentation to support this new setup.

### Transition to GitHub Codespaces:

* [`.devcontainer/codespaces-compose.yaml`](diffhunk://#diff-6925aac80a2f9e787209cae9f54f4c8f01e8e1249c54343a10ec9a9caf56cb8aR1-R32): Added a Docker Compose configuration for Codespaces, including services for the main application, MariaDB, and RabbitMQ.
* [`.devcontainer/devcontainer.json`](diffhunk://#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686L2-R27): Updated to use the new Docker Compose file, defined services, and set up port forwarding and commands for post-create and update content.
* [`.env.devcontainer`](diffhunk://#diff-6616b9ed7b02aaa0807761dd95de1eafd99bdd6c6a4b5ebb03925141badb3f24L2-R18): Updated environment variables to match the new Codespaces setup, including database and RabbitMQ URLs.
* [`docs/installation.md`](diffhunk://#diff-8b042e3f94ca5c59a7cd990b950aec0073ea84fe811e7b22be51158d7b180d56L5-R142): Revised documentation to guide users on setting up a development environment using GitHub Codespaces instead of Gitpod.

### Removal of Gitpod Configuration:

* `.gitpod.dockerfile` and `.gitpod.yml`: Removed Gitpod-specific Dockerfile and configuration as the project transitions to GitHub Codespaces. [[1]](diffhunk://#diff-e6428ece58f26fe63c0e7a9f12c2d4c98062c7982d291dab84b04b17b5554f3cL1-L28) [[2]](diffhunk://#diff-370a022e48cb18faf98122794ffc5ce775b2606b09a9d1f80b71333425ec078eL1-L84)

### Updates to Justfile:

* [`justfile`](diffhunk://#diff-deb9bb56fb122db0b605aa5b63f95a4665c905b18dd670e1fa6c877576a94ff1L1-R92): Updated commands to reflect the new environment setup and removed Gitpod-specific commands. Added new commands for setting up and running the development environment in GitHub Codespaces.


mrchrisadams:
For the _migrate set up command for main dev container to run migrations_ item above, we mainly want to take this part of the gitpod set up file:

```yaml

tasks:
  # set up python dependencies and database
  - name: main terminal
    init: |
      curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /home/gitpod/.local/bin/
      cp ./.env.gitpod ./.env
      python -m venv .venv
      source .venv/bin/activate
      python -m pip install --upgrade pip uv
      uv sync
      dotenv run -- ./manage.py migrate
      dotenv run -- ./manage.py tailwind install
      dotenv run -- ./manage.py tailwind build
      cd ./apps/theme/static_src/
      npx rollup --config
      cd ../../../
      dotenv run -- ./manage.py collectstatic --no-input
    command: ls
```

And turn it into something that works on start up with codespaces, by updating our `devcontainer.json` file with a post


```yaml
      # set up a copy for our env file
      cp ./.env.gitpod ./.env

      # probably convert these to "just" tasks in our just file, this lets us stop relying on dotenv
      # we will likely need to make sure we have a recent node runtime
      dotenv run -- ./manage.py migrate
      dotenv run -- ./manage.py tailwind install
      dotenv run -- ./manage.py tailwind build

      # turn this into a single task to generate the front end assets, then collect them 
      # into the correct place
      cd ./apps/theme/static_src/
      npx rollup --config
      cd ../../../
      dotenv run -- ./manage.py collectstatic --no-input
```

### Further reading

I was surprised how little documentation there was for this. The most helpful notes I came across Simon Willison's own notes on setting up a multi container codespace, but it's a bit old. 

https://til.simonwillison.net/github/django-postgresql-codespaces


The reason the Microsoft universal image is x86 only in the first places seems to be related to [oryx - something we don't use anyway](https://github.com/devcontainers/images/issues/545)

#### devcontainers

This link here lists the expected keys in the `devcontainer.json` spec. it helps us figure out what we use if we want a command to run on startup, like the gitpod `init` and `command` blocks.

https://containers.dev/implementors/json_reference/#image-specific

this link here provides a high level idea of how the devcontainers are supposed to run with docker compose:
https://containers.dev/implementors/spec/#docker-compose-based

this link here is github codespace's own documentation for using devcontainers in codespaces:
https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers

mrchrisadams:
OK, this was a lot of faff, as it turns out that gitpod does a lot of clever voodoo that is either complicated to replicate in Github codespaces, or as far as I could tell not really plausible. 

I've also taken steps to simply the justfile that we use, and remove files we no longer need, like the makefiles, and gitpod related files. We'll need to remove the gitpod extension associated with this repo, as i'm not sure when / how gitpod gets End of Lifed in April.

mrchrisadams:
@timcowlishaw this took quite a chunk of time to get it behaving, but I think this is in shape to review. Can you please review it? I'm happy to do on a synchronous pairing session.

.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,30 @@
...
+    "13306": {
+      "label": "MariaDB",
+      "onAutoForward": "notify"
+    },
+    "15672": {
+      "label": "RabbitMQ",
+      "onAutoForward": "notify"
+    }
+  },
+  "postCreateCommand": "cp .env.devcontainer .env && just dev_setup_codespaces || echo '.env.devcontainer not found'",
```
mrchrisadams, line 26: We have this defined in two places because we want to run this when we rebuild a devcontainer as well as when we create it for theh first time.

.env.devcontainer
```diff
@@ -1,21 +1,21 @@
...
 
 PYTHONDONTWRITEBYTECODE=1
 
 # drop into ipdb when breakpoint() is called
 PYTHONBREAKPOINT="ipdb.set_trace"
 
 SECRET_KEY='some-key'
-DATABASE_URL=mysql://root@127.0.0.1:3306/greencheck
-DATABASE_URL_READ_ONLY=mysql://root@127.0.0.1:3306/greencheck
+DATABASE_URL=mysql://root:deploy@db:3306/greencheck
```
mrchrisadams, line 10: for the devcontainer, can either:

1. have all the containers on the same network where they can access each other at 127.0.0.1:<PORT_NUMBER>, but not access them from the host machine, which means we can't use any GUI tools installed on the host, **_or..._**
2. use the default compose networking, where we do not connect to localhost from _inside_ the containers, but instead use the service names instead. This _does_ allow us to connect from 'the outside' so using a graphical tool like way... [Sequel Ace](https://sequel-ace.com/) for exploring databases, and other tools to speak exposed services works.

I've chosen the latter here.

.devcontainer/devcontainer.json
timcowlishaw: Just two things that might simplify this:

- Do we need the error check at the end? `.env.devcontainer` is tracked in git so should always be present, no?
- Could (and should?) we move the copying of the env file _into_ the just `dev_setup_codespaces` task so that each of the commands defined here can just call out to just, and all the details are kept in a single place?

If we do that (and possibly also if we don't), it might also be a good idea to have the command fail rather than overwrite an existing .env - give your personal .env isn't checked in and might have a bunch of hard-to-remember customisations, we don't want to make it too easy to overwrite accidentally.

docs/installation.md
```diff
@@ -2,80 +2,144 @@
...
+This project uses Github Codespaces to provide managed, hosted development environments on Github, using the [devcontainer.json](https://containers.dev/) standard. This means you don't need to install project dependencies on your local computer if you don't want to. 
 
-### Preparing a Gitpod workspace
+You can work in a browser or use a local editor like VS Code, or Vim to connect to the managed environment.
 
-After logging into Gitpod or creating an account we recommend also installing the Gitpod browser extension. This extension integrates extra functionality in websites such as Github.
 
-1. Go to our [Github repository](https://github.com/thegreenwebfoundation/admin-portal) and click on the Gitpod button if you have the browser extention installed or manually create and link it in your Gitpod Workspaces. It might take a couple of minutes to prepare the environment.
+### Setting up a Gitpod Codespace
```
timcowlishaw, line 16: a _Github_ codespace :-)

timcowlishaw (commented):
Fantastic, Chris, thanks! Have tested with @hanopcan also and she's very happy. Just a couple of small comments for your consideration but all looks good to me!

timcowlishaw (approved):

docs/installation.md
mrchrisadams: ah, thanks tim!

> Do we need the error check at the end? .env.devcontainer is tracked in git so should always be present, no?

We probably could remove it, _**but...**_ When I was working on this initial codespace set up I ended up running `gh codespace rebuild`, or selecting the "Codespaces: Rebuild Container" option quite a few times. This command is designed to allow you to run with uncommitted changes while testing out docker compose or devcontainer configurations before you check working config into source control.

We discussed you doing a bit of work to adapt this to use compose locally for local mariadb and rabbitmq services, and also possibly swapping in a dev image that works on bothx86 and arm64 machines.

Can I leave the decision to you to remove it to one that happnens outside this PR, when you've had a poke around with some docker buliding japery and then decided if it was useful for you? Seeing the value show up in the creation logs when debugging was helpful for me, but you're right , we might not need it any more.

> Could (and should?) we move the copying of the env file into the just dev_setup_codespaces task so that each of the commands defined here can just call out to just, and all the details are kept in a single place?

I would have liked this too, but I think just _**relies on the .env file being available**_ to populate an environment with the necessary environment variables for the just tasks to run. I think it would mean that the tasks ran without the required env vars set. I ended up using the `&&` option, because just doesn't preserve context from line to line (so a cd into one directory will not persist for the next line)

PR #647, opened 2025-04-01
Local development documentation + local devcontainers

timcowlishaw:
Add further documentation for developing locally, either with devpod, or using docker-compose to run service dependencies.

We also change the devcontainer base image to python, so that it can be run locally with devpod on Apple Silicon macs


timcowlishaw:
[Help get the platform running locally (review gitpod method)](https://trello.com/c/Zn678F09/520-help-get-the-platform-running-locally-review-gitpod-method)

timcowlishaw:
Need to give this a test on Github codespaces before merging, it should be fine, but just in case!

mrchrisadams:
Hi @timcowlishaw I took a look at this, and 

**With github codespaces**

This looks like it works as expected. I use github codespaces to open a VS Code environment, and when I try to set a machine, I get the main app running with the two services (Maria DB and RabbitMQ).

It takes a few minutes to provision the machines, then I have a running machine, with the `cp .env.devcontainer .env && just dev_setup_codespaces` steps run to set it all up. 

With the new slimmer image I seem to lose many of the nice developer experience things in the universal image we had before tho.

**With Devpod**

I had more trouble here, but got it working, and I now can spin up a local environment on my M1 Macbook using these same containers. So the arm64 blocker seems to be address.

When running the local devpod app I was first challenged to tell gitpod where to open the workspace folder.

Screenshot 2025-04-02 at 13 53 36

I was able to manually resolve the first time by telling vscode what files to mount liek the screenshot below

Screenshot 2025-04-02 at 13 53 41

**finding a fix for the what to mount' thing with Devpod**

I was able to resolve this by adding this proprerty to the `devcontainer.json` file

```json
"workspaceFolder": "/workspaces/admin-portal",
```

However, with devpod, I still don't see a way to run these commands:

```json
"postCreateCommand": "cp .env.devcontainer .env && just dev_setup_codespaces || echo '.env.devcontainer not found'",
"updateContentCommand": "cp .env.devcontainer .env && just dev_setup_codespaces || echo '.env.devcontainer not found'",
```

**Having separate devcontainer files - one for Github Codespaces, and one for Devpod**

In the end, I created a new file locally, `.devcontainer/devpod.devcontainer.json` for devpod to read, which had the `workspaceFolder` explicitly set. I could access this by opening the local folder with the checked out repo:

Screenshot 2025-04-02 at 16 16 23

...and telling devpod to load that file instead of the default devcontainer.json file:

Screenshot 2025-04-02 at 16 16 31

This gives me the locally set up environment via docker, but I can still make changes from the outside or use graphical Git commands and so on.

There still remains that awkward local issue to solve, but as we discussed, it seems to only affect `bash`, not `zsh`. I think it's solvable with some docker `locales` wrangling, but honestly in my case, starting with the zsh shell seemed to make them go away for me.

timcowlishaw:
Hey @mrchrisadams  - Have moved the "new" devcontainer.json to devpod.devcontainer.json and reverted the old "master" one, added the workspaceFolder directive to the local devpod config, and updated the documentation to reflect this (And address some of the other caveats above). Are we happy with this as a solution for now?

mrchrisadams (approved):
Hi Tim! I'm v happy with this - if we miss the universal dev image, we can always define it in the compose file to include it later. Feel free to merge in at your convenience.

timcowlishaw:
aaaaah no wait, i forgot that we define the image in the compose file, which complicates the "two seperate devcontainers" thing a bit - I  suspect i've just inadvertently broken the codespaces version (as it now would still use the python image, without adding any of the other dependencies we rely on which need to be in devcontainer.json). I think it's worth trying to get it working locally and on codespaces with the python image, if not i'll look at a way we can use a different image on codespaces vs devpods.

PR #648, opened 2025-04-08
Add carbon.txt badge in directory

timcowlishaw:
This commit adds a new carbon_txt_url field to hosting providers, which can be populated in the admin console, and a "carbon txt" badge to be shown on directory listings for those providers which have a carbon_txt url set. This allows us to highlight the presence of carbon.txt in the directory for providers that are using it, while side-stepping some of the issues pending resolution with domain verification for now.

## Badge UI: 
image

## Admin interface:

image

PR #649, opened 2025-04-08
Refactor accounts models

timcowlishaw:
That `hosting.py[ was getting quite unwieldy at over 1100 LOC, so i've had
a go at splitting it up: the external interface stays the same (as we
reexport everything used elsewhere in the codebase), but now the
"hosting" models are divided into a "datacentres" sub-package, a
"providers" sub-package, and an "abstract" sub-package for abstract
base classes shared by both. The "providers" package is still pretty
huge, at 700 LOC, but I think this makes things a little more
manageable, with room for further improvement.

@mrchrisadams, what do you think? This is meant as a bit of a
conversation-starter as much as anything - would you like to see some of
the larger files split up, or are you happy with everything in one
place?

### DIFF
--- a/apps/accounts/models/hosting/__init__.py
+++ b/apps/accounts/models/hosting/__init__.py
@@ -1,0 +1,69 @@
+from django.db import models
+from model_utils.models import TimeStampedModel
+
+from .abstract import EvidenceType, Label
+from .datacenter import (
+    Datacenter,
+    DatacenterCertificate,
+    DatacenterClassification,
+    DatacenterCooling,
+    DataCenterLocation,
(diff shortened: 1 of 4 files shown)


timcowlishaw:
(Note to self to check the model change in #648 gets included here after it's merged, and we rebase this branch, as I'm not sure git is that clever.)

mrchrisadams:
hi @timcowlishaw I'm v happy with this - thank you for proposing it, as I had found the file a bit unwieldy too.

I've rebased this from master after merging in #648, and once the tests pass, I'll merge it in.

PR #651, opened 2025-04-18
Eco ci update

ArneTR:
Eco CI, which was used in the `run-tests.yml` workflow was quite outdated.

This PR:
- Moved Eco CI in the `run-tests.yml` workflow to `@v4`. This will also auto-update.
  - In case dependabot is active on the repo and auto-updating is not wanted the version can also be pinned to `@a2305b330063eddbf6fd758bb5a335a87c65d804` which is the lastest version as of now
  - Dependabot will then open a PR in case we release a new version
- Also this PR introcuced Eco CI to the other workflows. Since they only have one signifcant step only one Eco CI step was integrated.

Note: 
In the current implementation Eco CI cannot derive CO2 values. An Electricitymaps-token is needed for that.
In case you either have a paid mult-zone-token or still have an "old" Electricitymaps token from the time where it was still called `CO2-Signal` you can use it. Newer "free" tokens will sadly not work as they are zone restricted :/  

A sample integration would look like this:
```
      - name: Eco CI Energy Estimation - Initialize
        uses: green-coding-solutions/eco-ci-energy-estimation@v4
        with:
          task: start-measurement
          electricitymaps-api-token: ${{ secrets.ELECTRICITYMAPS_TOKEN }}
```

=> A secret should be used for the token and must be accessible for the workflow

### DIFF
--- a/.github/workflows/run-deploy-staging.yml
+++ b/.github/workflows/run-deploy-staging.yml
@@ -20,6 +20,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+
+      - name: Eco CI Energy Estimation - Initialize
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4
+        with:
+          task: start-measurement
+        continue-on-error: true
+
       - name: Checkout code
         uses: actions/checkout@v4
 
@@ -69,3 +76,18 @@ jobs:
           AWS_CONFIG_FILE: ${{ secrets.AWS_CONFIG_FILE }}
           API_URL: ${{ secrets.API_URL }}
           TRELLO_REGISTRATION_EMAIL_TO_BOARD_ADDRESS: ${{ secrets.TRELLO_REGISTRATION_EMAIL_TO_BOARD_ADDRESS }}
+
+      - name: Eco CI Energy Estimation - Get Measurement
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4
+        with:
+          task: get-measurement
+          label: "ansible deploy staging"
+        continue-on-error: true
+
+      - name: Eco CI Energy Estimation - End Measurement
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4
+        with:
+          task: display-results
+          pr-comment: true
+          send-data: true
+        continue-on-error: true
--- a/.github/workflows/run-deploy.yml
+++ b/.github/workflows/run-deploy.yml
@@ -17,6 +17,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+
+      - name: Eco CI Energy Estimation - Initialize
+        uses: green-coding-solutions/eco-ci-energy-estimation@v4
+        with:
+          task: start-measurement
+        continue-on-error: true
+
       - name: Checkout code
         uses: actions/checkout@v4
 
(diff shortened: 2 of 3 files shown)


ArneTR:
I noticed the run failed, but it seems like not due to the inclusion of Eco CI.

Furthermore: The run just ran, although I am an external contributor. Is that intended?

timcowlishaw:
Hey there @ArneTR ! Aha, thanks so much for this - I'll do some further testing and see if I can't work out what that other failure is. Good question about running CI on external contributions - I'll defer to @mrchrisadams on that, but I suspect that might not be intended (and in fact ought to be changed)!

timcowlishaw:
The failure comes when `pytest` is called in the task at line 111 of `run-tests-yml`, and is caused by using boto3 to attempt to get the domain snapshot bucket::

```
  File "/home/runner/work/admin-portal/admin-portal/apps/greencheck/views.py", line 11, in <module>
    bucket = object_storage.green_domains_bucket()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/admin-portal/admin-portal/apps/greencheck/object_storage.py", line 29, in green_domains_bucket
    return object_storage_bucket(settings.DOMAIN_SNAPSHOT_BUCKET)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/admin-portal/admin-portal/apps/greencheck/object_storage.py", line 15, in object_storage_bucket
    object_storage = session.resource(
                     ^^^^^^^^^^^^^^^^^
```

... and the underlying error is:

```
   File "/home/runner/work/admin-portal/admin-portal/.venv/lib/python3.12/site-packages/botocore/endpoint.py", line 408, in create_endpoint
    raise ValueError(f"Invalid endpoint: {endpoint_url}")
```

This to me suggest that the problem is with the `OBJECT_STORAGE_ENDPOINT` setting, (object_storage.py, line 17), which in turn is taken from the environment variable OBJECT_STORAGE_ENDPOINT, which is populated in line 117 of `run-tests.yml` from the github secrets.

I don't have admin access to the repo so can't confirm that this is set (@mrchrisadams, when you're back from djangocon, could you take a look?), but given we haven't seen this failure on any other branches, I'm wondering if the new Eco CI update is somehow resetting the environment in some unexpected way, perhaps because of some weird interaction with dotenv (this is the only CI task which invokes dotenv, and also the only one failing)?

I can't quite see how this would happen, but can't think of any other possibilities at present.

ArneTR:
Maybe this might be from contributor PRs only.

We are seeing a similar behaviour, when our code relies on secrets, but for instance Dependabot opens a PR.

The pipeline runs, but the secrets will not be populated.

I have not investigated how the pipeline can be configured appropriately then but happy to copy if you find a solution.
If that is the reason however it is not related to Eco CI, but a general external contributor quirk.
When I look at your closed PRs all made by depenadbot also failed ... so maybe this is the case here.

Hope that input helps somehow ...

timcowlishaw:
Aha, thanks @ArneTR, that makes perfect sense - in fact we definitely *wouldn't* want PRs opened by external contributors to have access to secrets, as then they could fairly trivially add a job to `print(SECRET)` to the CI workflow, and it turns out that this is [widely known and documented behaviour](https://michaelheap.com/access-secrets-from-forks/). It looks like if we switch to manually approving PRs from forks before running, then we get the option to manually run the workflow (with secrets) after the code has been reviewed by a contributor. I suspect this is what we should do once @mrchrisadams is back - in the meantime I'll see if I can get this particular PR passing somehow!

timcowlishaw (dismissed):

timcowlishaw:
Actually, if i rerun having approved the PR, does that help?

mrchrisadams:
Hi @ArneTR  thanks so much for this, and sorry about the slow review. I've rebased the latest changes onto a version of this PR and run it in the PR below:

https://github.com/thegreenwebfoundation/admin-portal/pull/656

It's green, so I'm loading it in, and I'll chat with @timcowlishaw about manually approving PRs from forks next week.

Thanks again and have a nice long weekend!

ArneTR:
Sweet!

I see the runs are trickling in: https://metrics.green-coding.io/ci-index.html

In case you wonder why it shows no carbon and only energy: No electricitymaps token was configured. And sadly it will only really work if you have an old "CO2-Signal" token. The new free tier of them is zone locked and GitHub Machines tend to appear in different zones :/

Happy to get insights though if you know of any other vendors that provide free US CarbonIntensity data location based.

ArneTR:
Wow, good call.

I kinda feel stupid of not having thought about enabling people to just insert a static value :) Will implement that!

Using a fixed value as fallback I feel might be tricky as you get suprising spikes in your data ... having a NULL there would be cleaner I feel.
But these are details ...

Let me ping you once the static value option is in place with a PR. Thanks for the input!

PR #654, opened 2025-05-09
Add test to catch importing for archived providers

mrchrisadams:
This pull request introduces a new exception to handle cases where data import is attempted for an archived hosting provider, along with the necessary updates to enforce this behavior and tests to validate it. The changes ensure better error handling and maintain data integrity by preventing imports for archived providers.

### New exception for archived providers:
* Added `ImportingForArchivedProvider` exception in `apps/greencheck/exceptions.py` to handle cases where data import is attempted for an archived provider. This includes a default error message and code.

### Updates to data import logic:
* Modified `process_addresses` in `apps/greencheck/importers/network_importer.py` to check if the hosting provider is archived. If archived, it logs a warning and raises the new `ImportingForArchivedProvider` exception.
* Imported the new exception in `apps/greencheck/importers/network_importer.py`.

### Tests for archived provider handling:
* Added a new test, `test_handle_with_archived_provider`, in `apps/greencheck/tests/test_importer_amazon.py` to ensure that attempting to import data for an archived provider raises the `ImportingForArchivedProvider` exception.
* Added necessary imports for testing in `apps/greencheck/tests/test_importer_amazon.py`.

### DIFF
--- a/ansible/templates/import_ips_for_large_providers.sh.j2
+++ b/ansible/templates/import_ips_for_large_providers.sh.j2
@@ -11,6 +11,6 @@ cd {{ project_root }}/current/
 
 # run our ip imports
 source .venv/bin/activate
-#dotenv run -- ./manage.py update_networks_in_db_amazon
+dotenv run -- ./manage.py update_networks_in_db_amazon
 dotenv run -- ./manage.py update_networks_in_db_google
 dotenv run -- ./manage.py update_networks_in_db_microsoft
--- a/apps/greencheck/exceptions.py
+++ b/apps/greencheck/exceptions.py
@@ -20,3 +20,15 @@ class NoMatchingDomainHash(NotFound):
     An exception raised when we try to fetch a shared secret for a provider
     but no shared secret has been set.
     """
+
+
+class ImportingForArchivedProvider(Exception):
+    """
+    An exception raised when we try to import data for a provider
+    that has been archived.
+    """
+
+    default_detail = (
+        "This provider has been archived. Please unarchive it before importing data."
+    )
+    default_code = "archived_provider"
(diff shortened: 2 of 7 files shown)


timcowlishaw (commented):
This looks great to me, Chris - just one thing to talk through perhaps before merging - do we want the importer to fail early, or continue after logging a warning? We perhaps just need to check the current cron script behaviour - we don't want a check for the (archived) AWS provider to prevent the script importing (not archived) azure or Google cloud IP ranges if AWS is checked first!

mrchrisadams:
hi @timcowlishaw ! you raised a good point - we totally have this with the `set -euo pipefail` line in the script.

TBH, if we had to choose, I think we'd probably _would_ want the cronjob to survive one import being aborted on this ground, and at present the `set -euo pipefail` means it would fail early and noisily.

I think I should update this PR to have the management command catch the new exception, and log a sensible message to STDERR instead.

I'm less sure about what the status code should be though - it's still ok to to STDERR and return a `0` status code, right?

https://github.com/thegreenwebfoundation/admin-portal/blob/master/ansible/templates/import_ips_for_large_providers.sh.j2#L6C1-L16

timcowlishaw:
aha, that's a good question - I don't see any problem with logging to stderr and returning 0 -  [it appears to be pretty standard](https://unix.stackexchange.com/questions/594666/can-a-program-output-to-stderr-and-still-return-a-zero-exit-status).

I guess we _might_ want the overall script to return a non-zero status if *any* of the provider imports fail, but to attempt all providers regardless of whether one fails, although that feels a bit like bikeshedding - I think it's fine to return 0, especially as "not importing archived providers" is the defined behaviour, not something exceptional

mrchrisadams:
thanks @timcowlishaw that gives me the sanity check I think I needed to know how to finish this PR 👍

mrchrisadams:
@timcowlishaw when you get a mo, would you cast an eye over this, and if you're happy, merge it in please?

PR #655, opened 2025-05-12
Add 'basis for verification' question to provider request form

timcowlishaw:
This PR adds a new, compulsory "what basis are you claiming for green web verification?" question to the first page of the provider request wizard, as a form of triage - in order to reduce the number of unsuccessful requests that @fershad has to deal with (see [This trello ticket](https://trello.com/c/WZgVt1At/438-add-a-what-is-your-basis-for-verification-question-to-the-verification-flow)).

The possible responses here are stored in a new Taggit model in the database, so are editable through the admin portal, and the given responses on a given providerrequest are also visible in the admin. The question shows on the first page of the wizard, and the responses are shown in the final confirmation / preview step.

Question in wizard:
image
Show in preview / confirmation page at end of wizard:
image
Accessible from admin:
image


timcowlishaw:
[Add a "What is your basis for verification?" question to the verification flow.](https://trello.com/c/WZgVt1At/438-add-a-what-is-your-basis-for-verification-question-to-the-verification-flow)

fershad:
@timcowlishaw could we merge this into Staging first so that I can step through it from both the verification wizard & admin panel perspective

timcowlishaw:
Updated to add "more info" links and to copy verification basis to hosting provider on approval!

timcowlishaw:
@fershad  - I'll get this up onto staging today, just want to check something wiith @mrchrisadams first!

apps/accounts/migrations/0068_verification_basis_permissions.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("accounts", "0067_verificationbasis_and_more"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            group_permissions_2025_05_12_add_verification_basis_permissions_for_admins
```
mrchrisadams, line 14: One note - I started taking this approach a while back, and I'm not sure if it's the idiomatic way to handle migrations, but it seemed to work. I'm open to using a more django-y approach if it seems weird.

mrchrisadams (approved):
Hi @timcowlishaw  looks good! Only a few small niggles, and no show stoppers. Danke!

apps/accounts/migrations/0068_verification_basis_permissions.py
timcowlishaw: just commenting here for future reference: we had a chat about this yesterday, and have created a new card for a more general review of how we're handling permissions and data migrations! the current approach makes sense to me but doesn't quite seem to be doing what we expect.

Issue #631, opened 2025-02-24
Introduce a "time to live" concept for green domains


mrchrisadams:
At present, our GreenDomains table has a `modified` property, and we can clear a cached lookup by passing in a `no-cache` param to a green web check, when we suspect a cache result might be out of date / stale.

Once the no-cache check has been run, the result is then queued up to either delete the green domain if the result is now gray, OR update the result if it still shows as green.

The thing is, for the majority of our checks we don't do these slow no-cache style checks - we follow the route to check the cache table first. this makes the result respond quickly, and it also triggers a worker to log the updated check.

Ideally our system would track when a cached result was _created_, rather than just _modified_, and we'd be able to run a `no-cache` style check after a domain existing in the green domains table for a given amount of time - this would clear out the old/stale domain check results we sometimes get.

mrchrisadams:
**Note:** there's some internal notes in our Trello instance below that provides extra context

https://trello.com/c/kFxvqNQY/498-introduce-a-time-to-live-concept-for-green-domains#comment-67bc44422d2d663801301d51

PR #658, opened 2025-06-13
Actively purge greendomains for archived providers, and introduce a created date field

timcowlishaw:
A first step towards a solution to #631  - there are nearly 2 million rows in the greendomains table which are associated with archived providers, and are therefore not actually green, which is both bloating the table (And causing the dataset export to fail) and means that the data provided is innacurate.

This PR does four things:

1) Ensures greendomains are proactively deleted when a provider is archived, in order that the dataset is kept up to date
2) Provides a management command to do a one-off purge of existing archived domains
3) adds a `created` datetime field on greendomains, defaulting to the existing `modified`date for existing rows, and set on creation for new ones, which will help us in future if we want to more aggressively prune thsi table
4) Amends the dataset export script to use the database port set in django settings, which ensures it can be used locally (and the tests pass)

When deploying, four things need to happen, in this particular order:

1) First, a **dump of the greendomains table should be taken**, against the production DB; from a machine with sufficient RAM for it to run successfully. This dump should be downloaded and verified before proceeding: ` just dev_manage dump_green_domains --upload`
2) Second, **migrations  should be run to add the created flag** - this LOCKS the greendomains table and runs for several minutes, so worth deploying this outside of peak hours, and ideally with some kind of "maintenance mode" holding page (aside from the annoyance to users, we don't want to flood Sentry with errors again)
3) Only then should `just dev_manage clear_archived_greendomains` be run, to clear the existing archived provider domains.
4) Finally,  `just dev_manage dump_green_domains` should be run again, on app1 this time but without the `--upload` flag, and we should verify that the database dump is generated successfully.

### DIFF
--- a/apps/accounts/models/hosting/provider.py
+++ b/apps/accounts/models/hosting/provider.py
@@ -437,6 +437,9 @@ class Hostingprovider(models.Model):
         active_green_asns.update(active=False)
         active_linked_domains.update(active=False)
 
+        # When providers are archived, any domains they host cease to be green, so we should
+        # Remove them from the cache of green domains. This ensures that the next time the are
+        # checked, the correct (grey) result will be returned.
         from apps.greencheck.models import GreenDomain # Avoid circular import
         GreenDomain.objects.filter(hosted_by_id=self.id).delete()
 
--- a/apps/greencheck/management/commands/clear_archived_greendomains.py
+++ b/apps/greencheck/management/commands/clear_archived_greendomains.py
@@ -5,7 +5,7 @@ from ...models import GreenDomain
 
 
 class Command(BaseCommand):
-    help = "Clear cached GreenDomain records for providers that are archived)"
+    help = "Clear cached GreenDomain records for providers that are archived"
 
     def handle(self, *args, **options) -> None:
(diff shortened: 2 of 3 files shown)


apps/greencheck/migrations/0021_add_created_to_greendomains.py
```diff
@@ -0,0 +1,29 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="greendomain",
+            name="created",
+            field=models.DateTimeField(auto_now_add=True, null=True),
+            preserve_default=False,
+        ),
+        migrations.RunSQL(
+            sql="UPDATE greendomain SET created=modified",
```
mrchrisadams, line 21: Hey @timcowlishaw  I've checked locally and it looks like with MariaDB, this migration will acquire a table level lock, when making this query.

On my machine, the migration locked the green domains table for nearly 2 minutes.

This is ok, it's better to be pre-warned when it happens.

timcowlishaw:
Hey @mrchrisadams  - would you like me to do the merge on this one? the underlying error it's fixing is now ever so slightly more visible, as the dataset browser is updating automatically again (and therefore loads an empty dataset currently, which has to be fixed by hand)

PR #662, opened 2025-06-19
Add a cron job to clear archived domains

timcowlishaw:
one missing piece of the jigsaw in #658 was a cronjob to actually run the "archived greendomain purger" task periodically. This corrects that oversight :-)


timcowlishaw:
@mrchrisadams another minor "FYI" here - will merge directly as it's small.

PR #664, opened 2025-06-20
Update copy for linking domains

fershad:
This PR makes only text updates to the various pages that have been added to facilitate linking domains to hosting providers.

All the changes have been made to the user facing wizard & details pages. Not changes have been made to the internal platform pages.

### DIFF
--- a/apps/accounts/templates/provider_portal/provider_domain_new/base.html
+++ b/apps/accounts/templates/provider_portal/provider_domain_new/base.html
@@ -25,8 +25,14 @@
       <div class="prose mx-auto mt-8">
         {% block explanatory_copy %}
             <p>Carbon.txt is a single recognisable location on any web domain for public sustainability data relating to that company. By combing Carbon.txt with your Green Web Foundation provider listing you can maintain a single source of truth for your organisations sustainability disclosures, and then share them across the different domains your organisation owns. This benefit can also be passed on to your customers, who's domains can also be linked to your hosting provider listing our dataset. This ensures that whenever their domains are checked in our <a href="https://www.thegreenwebfoundation.org/green-web-check/" target="_blank">Green Web Check</a> and <a href="https://admin.thegreenwebfoundation.org/api-docs/" target="_blank">API</a> the results accurately reflect their use of your services.</p>
-            
-            <p>More copy explaining what is required (access to server, DNS, etc), and how long the process will take and what it will involve</p>
+
+            <p>To continue, you will need either:</p>
+
+            <ul>
+                <li>The ability to upload a Carbon.txt file to your website server. Or,</li>
+                <li>The ability to make DNS updates for the domain you wish to link to your provider. Or,</li>
+                <li>The ability to edit HTTP Headers for the domain you wish to link to your provider.</li>
+            </ul>
         {% endblock explanatory_copy %}
         {%  block previous_steps %}
         {%  endblock previous_steps %}
--- a/apps/accounts/templates/provider_portal/provider_domain_new/base.html
+++ b/apps/accounts/templates/provider_portal/provider_domain_new/base.html
@@ -13,10 +13,10 @@
         <p class="mb-6 text-sm text-green"><button name="wizard_goto_step" type="submit" value="{{wizard.steps.prev}}" class="text-neutral-500 underline hover:no-underline">&lt; Back to previous step</button></p>
         {% endblock back_link %}
 
-        <h1 class="text-6xl mb-0"><span class="uppercase">Link a domain with carbon.txt</h1>
-        <h2 class="mt-4">
+        <h1 class="text-center">Link a domain with carbon.txt</h1>
+        <h3 class="my-3">
           Provider: {{provider.name}}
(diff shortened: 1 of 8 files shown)


apps/accounts/templates/provider_portal/provider_domain_new/base.html
```diff
@@ -24,8 +24,17 @@ <h2 class="mt-4">
...
-            <p>More copy explaining what is required (access to server, DNS, etc), and how long the process will take and what it will involve</p>
+            <p>Carbon.txt is a single recognisable location on any web domain for public sustainability data relating to that company. By combing Carbon.txt with your Green Web Foundation provider listing you can maintain a single source of truth for your organisations sustainability disclosures, and then share them across the different domains your organisation owns.</p>
+            <p>This benefit can also be passed on to your customers, who's domains can also be linked to your hosting provider listing our dataset. This ensures that whenever their domains are checked in our <a href="https://www.thegreenwebfoundation.org/green-web-check/" target="_blank">Green Web Check</a> and <a href="https://admin.thegreenwebfoundation.org/api-docs/" target="_blank">API</a> the results accurately reflect their use of your services.</p>
+
+            <p>To continue, you will need either:</p>
```
timcowlishaw, line 30: Just one tiny thing here, i think we need an `{% if provider.primary_linked_domain %}` conditional here, which displays the text as written *IF* a primary domain is set, and just "The ability to upload a Carbon.txt file to your website server." otherwise? The validator won't complain if you point your primary domain at a domain with A DNS TXT record or HTTP header instead of an actual self-hosted carbon.txt, but it's not really a real use case, and we maybe want to be a bit more explicit about that

apps/accounts/templates/provider_portal/provider_domain_new/base.html
```diff
@@ -24,8 +24,17 @@ <h2 class="mt-4">
...
+
+            <ul>
+                <li>The ability to upload a Carbon.txt file to your website server. Or,</li>
+                <li>The ability to make DNS updates for the domain you wish to link to your provider. Or,</li>
+                <li>The ability to edit HTTP Headers for the domain you wish to link to your provider.</li>
+            </ul>
```
timcowlishaw, line 36: I guess we're kinda glossing over the need to "have the information to hand to actually create the carbon.txt", but that's a bit of a can of worms that perhaps we should open at a later date

apps/accounts/templates/provider_portal/provider_domain_new/step_1.html
```diff
@@ -7,49 +7,62 @@ <h3 class="border-none">Domain: </h3>
...
       <li class="pl-0 m-0 grow">
-          <a href="#" role="tab" aria-selected="true" aria-controls="carbontxt-file-panel" id="carbontxt-file-tab" class="tab inline-block p-4 border rounded-tl-lg w-full active m-0 no-underline hover:underline">Create a new carbon.txt</a>
+          <a href="#" role="tab" aria-selected="true" aria-controls="carbontxt-file-panel" id="carbontxt-file-tab" class="tab inline-block p-4 border rounded-tl-lg w-full active m-0 no-underline hover:underline">Create a new Carbon.txt file</a>
       </li>
       <li class="pl-0 m-0 grow">
-          <a href="#" role="tab" aria-selected="false" aria-controls="carbontxt-dns-panel" id="carbontxt-dns-tab" class="tab inline-block p-4 border-y bg-neutral-200 hover:bg-neutral-100 w-full m-0 no-underline hover:underline">Use existing with DNS</a>
+          <a href="#" role="tab" aria-selected="false" aria-controls="carbontxt-dns-panel" id="carbontxt-dns-tab" class="tab inline-block p-4 border-y bg-neutral-200 hover:bg-neutral-100 w-full m-0 no-underline hover:underline">Use existing Carbon.txt with DNS</a>
```
timcowlishaw, line 31: Hmm, the extra width introduced here has slightly broken the layout of that tabbed box for me: 
image
(Firefox, 1920x1080) - I can take a look at this if needed.

apps/accounts/templates/provider_portal/provider_domain_new/base.html
fershad: I was thinking about this as well. I did have that gate in there previously, but removed it.
It does make sense that one primary domain _should_ always have its own carbon.txt file though.

For this to be complete, we would:

1. Need to put a gate here to check if the primary domain already exists.
2. Need to update the previous step (where the user inputs the domain). That step should:
    1. Check if a primary domain already exists for the provider.
    2. **If so** - show all the options as they currently are.
    3. **If not** - hide the option for the user to select the kind of domain they are linking & set the input domain as their primary domain by default (probably a hidden form field on the page).

timcowlishaw:
Thanks folks! @hanopcan @fershad I'm gonna review this with @mrchrisadams shortly, and I think we'll go ahead and merge and I'll do the last few changes on a new branch. It won't make any public-facing difference as all this stuff is behind a feature flag anyway, and will mean we can also merge in #666  and #667 .

PR #666, opened 2025-06-23
Tc refactor forms, views, and admin

timcowlishaw:
Some more basic "boy scout rule" stuff - breaking up the (very large) admin.py, views.py and forms.py files in accounts into smaller, meaningful units. This should make the codebase a little easier to navigate and understand.

Leaving it draft til #664 is merged, to avoid giving @fershad any merge headaches!

PR #667, opened 2025-06-23
Fix pending migrations

timcowlishaw:
We've had a bunch of migrations automatically generated with `makemigrations` for a while that we've never run, as we were worried they might be unintentionally destructive of data: due to the fact that MariaDB enums rely on the order in which values are defined, changing the possible values of an enum might change the actual stored values, if the new value set doesn't maintain the same order.

I've tested this and verified that no data is changed (verifying that the order of enum definitions remain the same, and that a `select ENUM_FIELD count(id) group by ENUM_FIELD` returns the same totals before and after the migration).

On deployment, it'd be worth:

- **taking a database backup before running the migrations**
- **running the same tests**: `select type, count(id) from greencheck group
  by type order by count(id) desc;` and `select type, count(id) from greencheck group by type order by count(id) desc;` before and after running the migration, and ensuring that the totals stay the same.

  Once this is run, the django database schema, django model
  definitions, and database schema should all be in sync, and `makemigrations` will not create masses of spurious changes everytime we run it.


timcowlishaw:
Leaving this as draft until the two previous PRs it's based off are merged (#664 and #666)

apps/greencheck/migrations/0022_delete_dailystat_delete_greencheckstats_and_more.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("greencheck", "0021_add_created_to_greendomains"),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
```
timcowlishaw, line 14: *if* we want to keep these tables around for any reason, we should wrap these in a SeparateDatabaseFromState instance to update the schema without deleting the data. To discuss with @mrchrisadams

timcowlishaw:
[on call with Chris] We're happy to deploy this (with the checks and backups described above), but again, tomorrow, when Tim's around for longer to deal with any unforeseen consequences (if any)

apps/greencheck/migrations/0022_delete_dailystat_delete_greencheckstats_and_more.py
timcowlishaw: We're happy these can go - we've got backups!

PR #668, opened 2025-06-27
Further Linked Domains UX improvements

timcowlishaw:
-  Link to the validator from  the domain details page - I think it makes sense to do this from the details page rather than the provider domains listing, so as not to make that list too cluttered. I’m linking to the validator (For now), as we don’t currently store the location of the carbon.txt itself within the platform - but we can validate the domain itself. Once I’ve made the change in the validator API to send back the discovered carbon.txt URL, ,I can then update admin portal to also show a link directly to the carbon.txt.
 - Prepopulate domain address with the provider’s own domain when no primary domain yet exists. I’ve also made sure that “secondary domain” is selected by default when a primary domain does exist, so the form should always “do the right thing” by default while allowing users to override as needed.
- Provided an example in “help text” for the domain field to show the format we’re expecting, which I think is enough of a hint. Explaining the format comprehensively (As in the error message) is a bit of a “wall of text” and is quite difficult to parse the meaning of, so using an example (along with the more detailed explanation in the error message) allows for a sort of progressive-revealing of more detail.
 - Add an anchor tag on the “Step 2” heading on the second form step, and link directly to it on submitting “Step 1” - this helps make the transition between steps more obvious.

    Add additional messaging both before and after submitting the domain for validation to make it clear that manual validation will take place, set expectations about timescales, and thank the user for their time and effort

### DIFF
--- a/apps/accounts/forms/linked_domains.py
+++ b/apps/accounts/forms/linked_domains.py
@@ -11,7 +11,7 @@ class LinkedDomainFormStep0(forms.ModelForm):
 
     domain = forms.CharField(
         help_text=(
-            "Enter a valid domain name, for example: <strong>thegreenwebfoundation.org</strong>"
+            "Enter a domain name, for example: <strong>thegreenwebfoundation.org</strong>, without the https://."
         )
     )
 
--- a/apps/accounts/templates/provider_portal/provider_domain_new/base.html
+++ b/apps/accounts/templates/provider_portal/provider_domain_new/base.html
@@ -24,13 +24,13 @@
     <article>
       <div class="prose mx-auto mt-8">
         {% block explanatory_copy %}
-            <!-- <p>Carbon.txt is a single recognisable location on any web domain for public sustainability data relating to that company. By combing Carbon.txt with your Green Web Foundation provider listing you can maintain a single source of truth for your organisations sustainability disclosures, and then share them across the different domains your organisation owns.</p>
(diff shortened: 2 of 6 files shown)


mrchrisadams:
@timcowlishaw  - I'm assuming you don't need me to review this until you've had a chance to reviews things on Thursday with Han.

This is currently assigned to me to review, so can you tag me when you'd like me to look it over?

timcowlishaw:
hey Hey Chris!

Aha, this can probably get reviewed/ merged actually, I'm going to start a
new branch for the next round of changes as it'll be a more fundamental
rethink. I still think there's value in merging this in though as there's
stuff in here we will still need!

mrchrisadams:
hey @timcowlishaw  - I see review is assigned to me in this PR, but it's not obvious from the UI when I ought to review it, as this is fairly long lived PR.

Would you mind explicitly  `@` mentioning  me when when you'd like me to have a detailed look? I think I have time tomorrow before I go on leave.

timcowlishaw:
Heya @mrchrisadams, aha, I was just about to message you - I actually don't think there's anything to do here, as it's all basically superseded by the new design, and also behind a feature flag. I'm going to merge this so we have it in the main git history, in case we want to rethink the new "linked domains" plan in future, but I don't think it needs particularly detailed review, it's all mostly cosmetic, and explicitly flagged, so will probably never see the light of day in its current form anyway!

PR #669, opened 2025-07-08
Add `listed_provider` flag to Greencheck API

timcowlishaw:
This PR adds a `listed_provider` flag to results from the greencheck API, so that we can hide the link to the directory in the green web checker UI when the provider is not listed.

The `listed_provider` status comes from what was previously known as the `showonsite` flag on a hosted provider, so i have renamed this to `is_listed` for consistency, and to make its meaning more clear.

There's also another sublety: we need to denormalise this flag onto the greendomains table, so we also need to invalidate the greendomains cache for a given hosting provider when its `is_listed`status (along with its `website` or `name`) changes, to ensure consistency - I've added logic to do this.

Finally, I've ensured that the `is_listed` status is editable by admins.

### DIFF
--- a/apps/accounts/models/hosting/provider.py
+++ b/apps/accounts/models/hosting/provider.py
@@ -554,12 +554,14 @@ class Hostingprovider(models.Model, DirtyFieldsMixin):
         )
 
     def save(self, *args, **kwargs):
+        # The is_listed flag, name and website url are denormalized into the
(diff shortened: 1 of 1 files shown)


mrchrisadams:
Thanks for this @timcowlishaw - we make some changes the database here, so would you mind running this on staging, and run the migration there, before we merge in and deploy to production? Otherwise, please deploy at your convenience

PR #681, opened 2025-08-13
Allow providers to link and validate their carbon.txt, and use it in the greencheck process.

timcowlishaw:
As set out in the "Linked Domains v2.5" miro board.

- Linked Domains are now gone, replaced by a CarbonTxt model which represents the existence of a carbon.txt for a provider and tracks the state of the carbon.txt setup process.

- Providers also have "carbon.txt motivation" tags, which store the responses to the "survey question" about problems they're experiencing in carbon.txt onboarding.

- Carbon.txt lookups are now done at the point of _check_ rather than in advance. the CARBON_TXT_RESOLUTION_TIMEOUT env variable will have to be tuned based on load in production to ensure this doesn't become a bottleneck.


apps/greencheck/domain_check.py
```diff
@@ -265,9 +250,8 @@ def check_domain(self, domain: str) -> SiteCheck:
         """
         UNRESOLVED_ADDRESS = "0.0.0.0"
```
timcowlishaw, line 252: This is where we would add the cache lookup @mrchrisadams

apps/accounts/models/hosting/carbon_txt.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Method to identify if a valid provider carbon.txt exists for a given domain.
+        First it performs a carbon.txt delegation lookup, attempting to resolve a canonical
+        carbon.txt location for the domain. If this is found, it then checks for a
+        ProviderCarbonTxt in our database corresponding to that url.
+        In the case where a provider carbontxt exists, it is returned, otherwise
+        returns None.
+        """
+
+        finder = FileFinder(http_timeout=settings.CARBON_TXT_RESOLUTION_TIMEOUT)
```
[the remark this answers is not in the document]
timcowlishaw: image

apps/accounts/migrations/0083_carbontxtmotivation_providercarbontxt_and_more.py
```diff
@@ -0,0 +1,185 @@
...
+                    created, modified, domain, created_by_id, provider_id,
+                    TRUE as is_delegation_set,
+                    CONCAT('https://', domain, '/carbon.txt') as carbon_txt_url
+                FROM accounts_linkeddomain
+                WHERE is_primary IS TRUE
+
+            """,
+        ),
+        migrations.RunPython(set_carbon_txt_urls),
+        migrations.DeleteModel(
```
timcowlishaw, line 182: Move this to a later migration in a PR when we've validated everything works (along with changes to the models)

PR #684, opened 2025-09-03
Deploy new app2a worker box.

timcowlishaw:
This runs ubuntu 24.04 which required a few changes in Ansible.

In particular: 

- the libmariadbclient-dev library no longer exists (it's included in libmariadb-dev)
- uv is now installed in .local/bin
- for some reason ansible shell invocations no longer load .bashrc, so we have to ensure we correctly invoke fnm in tasks that use node.js

### DIFF
--- a/ansible/_add_system_dependencies.yml
+++ b/ansible/_add_system_dependencies.yml
@@ -24,7 +24,7 @@
     pkg:
       - build-essential
       - libmariadb-dev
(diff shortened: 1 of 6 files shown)


mrchrisadams (approved):
Hi @timcowlishaw - thanks for this. I think it would be good to find a better solution to requiring all the FNM shenanigans, but it does work. Let's merge it in.

PR #685, opened 2025-09-23
Delete old linked domains table

timcowlishaw:
...which is replaced by the new ProviderCarbonTxt model, which is now fully deployed.


timcowlishaw:
As discussed Monday, I'm gonna merge this in, as the new carbon.txt implementation is working without a hitch!

PR #745, opened 2026-05-08
API Keys - backend

timcowlishaw:
This uses djangorestframework-api-key to implement API key access, and
adds a private key introspection endpoint for use in other GWF services.

Keys can have an attached note, and an optional expiry date.

It also includes an admin page for API keys, accessible to admin users only.

This requires two environment variables to be set:

- MAX_API_KEYS_PER_USER for the total number of valid API keys a user may have at one time.
- GWF_SHARED_SECRET is a shared secret for authentication to the
  introspection endpoint.


apps/accounts/models/api_key.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_usable_keys(self):
+        return self.filter(
+            models.Q(revoked=False),
+            models.Q(expiry_date__isnull=True) | models.Q(expiry_date__gt=timezone.now())
+        )
+
+    def create_key_for_user(self, user, note="", expiry_date=None):
+        if self.get_usable_keys().filter(user=user).count() >= settings.MAX_API_KEYS_PER_USER:
+            raise ValueError("Only 3 active API keys allowed per account.")
```
mrchrisadams: I think we want this to be `settings.MAX_API_KEYS_PER_USER` not 3.

ValueError(f"Only {settings.MAX_API_KEYS_PER_USER} active API keys allowed per account.")

mrchrisadams (approved):
Hi tim! Only one small nitpick.

Thank you!

mrchrisadams (approved):

PR #746, opened 2026-05-15
Implement support for linked providers in the admin platform

mrchrisadams:
See trello card 504
https://trello.com/c/cxLe0qC1/504-allow-providers-to-link-to-their-upstreams-in-verification-requests

> Providers who select Using an existing verified provider as the basis of verification can select one/more active verified provider from a pre-filled list. This update should include text which makes it clear that the disclosure of this relationship may be publicly displayed and shared by the Green Web Foundation through our online platforms and tools.

Key question:

- [x] confirm if filtering linked providers choices to the same country is sensible. It will likely stop people using providers active in multiple regions


mrchrisadams:
@timcowlishaw this PR is almost finished - I mainly need confirmation about whether we indeed filter possible providers to the same country, or not, but I think there is enough there to make it worth getting a second pair of eyes on it before merging in.

apps/accounts/forms/provider_request_wizard.py
```diff
@@ -140,29 +151,75 @@ class BasisForVerificationForm(forms.ModelForm):
...
+            "<p>When you are ready to continue, please select one or more of the statements below that apply to your organisation.</p>"
+        ),
+    )
+
+    linked_providers = forms.ModelMultipleChoiceField(
+        queryset=Hostingprovider.objects.filter(archived=False, is_listed=True),
+        required=False,
+        label="Which existing verified provider(s) do you rely on as the basis the claim for using green energy?",
+        help_text=mark_safe(
+            "[TODO: INSERT GUIDANCE ABOUT GREEN PROVIDER BEING VISIBLE] "
```
timcowlishaw: Just leaving a comment here to make sure this doesn't get forgotten :-)

apps/accounts/models/hosting/provider.py
```diff
@@ -228,8 +236,18 @@ class Hostingprovider(models.Model, DirtyFieldsMixin):
         blank=True,
         related_name="labels",
     )
-    is_listed = models.BooleanField(verbose_name="List this provider in the greenweb directory?", default=False)
+    is_listed = models.BooleanField(
+        verbose_name="List this provider in the greenweb directory?", default=False
+    )
     website = models.URLField(max_length=255)
+    linked_providers = models.ManyToManyField(
```
timcowlishaw: This is possibly a minor thing, (and possibly a bit of a pain to change), but I'm a little confused by some of the language here - we have linked_providers (which we refer to a upstream providers  in UI copy), and linked_by_providers (which we also call downstream providers elsewhere).

I think i'd favour renaming this relation to upstream_providers, and making the related_name downstream_providers, and standardising on that everywhere, as it'd make the whole thing much more understandable - we only need to remember which way the "stream" flows.

In fact, even better, might be something like client_of_providers (with related_name supplier_to_providers) to make the whole thing much more explicit

apps/accounts/templates/provider_registration/basis_for_verification.html
```diff
@@ -1,6 +1,61 @@
...
-</section>
+    <section class="prose mt-10">
+        <h1 class="mb-0">Basis for verification</h1>
+    
+        {% flag "linked_providers" %}
+        <div class="alert__warning mt-6" id="linked-providers-disclosure">
+            <p class="font-bold">Linked providers</p>
+            <p>
+                If your organisation relies on one or more already verified green
+                providers, you can list them below. [TODO: INSERT GUIDANCE ABOUT
```
timcowlishaw: Just flagging another TODO in copy

apps/accounts/views/provider/request/wizard.py
```diff
@@ -530,11 +564,15 @@ def _network_footprint_initial_data(hosting_provider: Hostingprovider):
             network_dict = {
                 "ips": [
                     {"start": range.ip_start, "end": range.ip_end}
-                    for range in hosting_provider.greencheckip_set.filter(active=True).all()
+                    for range in hosting_provider.greencheckip_set.filter(
```
timcowlishaw, line 558: Note for both of us for the future, as I'm also guilty of this , but when there's a huge amount of autoformat changes all over the codebase like this, in future it might be good to commit them as a separate PR as it makes the substantive changeset a bit difficult to follow

timcowlishaw (commented):
In general looks good, Chris - my main concern is about terminology. See the comment below! Thank you!

apps/accounts/views/provider/request/wizard.py
mrchrisadams: Yeah, that's a good point. I've created this to get it sorted once and then just keep to it in future, using the editor support that's avaiable

https://github.com/thegreenwebfoundation/admin-portal/issues/749

apps/accounts/models/hosting/provider.py
mrchrisadams: This is a good shout, thanks Tim. We have used upstream and downstream before when discussing this, and I'll follow your suggestion there.

PR #747, opened 2026-05-21
UI for API key management

timcowlishaw:
- Allows users to create and revoke their own keys, and adds ability for admins to ban API access for specific users and to raise the API key limit on a per-user basis.

  Users must answer a "why do you need API access" question before accessing the API keys section, but manual approval of their response is not required. The answer given is available in the admin UI.

  All of this is behind the "api_keys" feature flag, and this PR requires migrations to be run on deploy.


apps/accounts/forms/api_keys.py
```diff
@@ -0,0 +1,38 @@
+from django import forms
+
+from ..models import APIKey
+
+
+class APIAccessForm(forms.Form):
+    api_access_motivation = forms.CharField(
+        widget=forms.Textarea(), required=True,
+        label="Why do you need an API key?",
```
mrchrisadams: Tim, can you reframe this to sound a little bit less confrontational? 

Perhaps something like:

> What is your intended use case? 

Or simply the label being _'intended use case'_ instead?

timcowlishaw: Haha, yes, of course! :-D

PR #748, opened 2026-05-21
Remove old deprecated API token code

timcowlishaw:
This removes the old apitokens support (not actually used for auth anywhere).

Some users have tokens, see [this database table](https://admin.thegreenwebfoundation.org/explorer/41/), and we get very infrequent requests to the api-token-auth endpoint (see [grafana](https://grafana.greenweb.org/explore?schemaVersion=1&panes=%7B%22wjq%22:%7B%22datasource%22:%22H0yiLpF7z%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bunit%3D%5C%22web_prod.service%5C%22%7D%20%7C%3D%20%60api-token-auth%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22H0yiLpF7z%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now%2Fy%22,%22to%22:%22now%2Fy%22%7D%7D%7D&orgId=1)). Therefore we should contact these users before merging to advise them that this is going away.

### DIFF
--- a/apps/greencheck/tests/test_api.py
+++ b/apps/greencheck/tests/test_api.py
@@ -4,7 +4,6 @@ import pytest
 from corsheaders.middleware import ACCESS_CONTROL_ALLOW_ORIGIN
 from django.contrib.auth import get_user_model
 from django.urls import reverse
-from rest_framework.authtoken import models, views
 from rest_framework.test import APIRequestFactory
 
 from ...accounts import models as ac_models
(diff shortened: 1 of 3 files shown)


timcowlishaw:
Moving this to Draft - we won't merge until we're ready to inform the users of the old tokens that they're no longer supported.

timcowlishaw:
This is going to be a pain to merge on top of #751, and on top of that we don't know whether we're actually going to want to do this any time soon, so closing.

PR #751, opened 2026-06-02
Edits to linked providers copy

fershad:
This PR:

- makes some text changes to the provider linking copy on the "basis for verification" page.
- repositions the alert box notifying that linked providers may be shared publicly.
- fixes layout issues with linked providers on the summary page.
- adds a banner to the home page of the portal notifying about the upcoming verification changes.


mrchrisadams (approved):
@fershad  and I discussed this on a video call ust now, after I got the last  test passing again.

This is in good enough shape to get to production, and the markup tweaks we discussed we will address separately as they're not blockers to rolling out the feature.

PR #752, opened 2026-06-05
Fix 500's when visiting pages guarded by is_admin

mrchrisadams:
At the moment, visiting some paths like the SQL explorer path used by our staff , and Some user facing paths like the provide a request update wizard ones can trigger 500 errors, because we often check for `is_admin` when we do some of our access checks. 

The default anonymous user does not have this method, and trying to call it raises an exception, triggering the 500. 

This change introduces a more graceful fallback, and verifies that the pages that were throwing errors no longer throw errors. 

Machine generated summary below

----

This pull request addresses issues where code was assuming the presence of an is_admin property on Django's AnonymousUser, which previously caused AttributeError exceptions for unauthenticated users. The main change is a patch to AnonymousUser to ensure is_admin always exists and returns False, preventing errors across the application. Additional tests have been added to verify that unauthenticated users do not trigger server errors when accessing certain endpoints.

Key changes:

**AnonymousUser patching:**
* Added a patch in `apps/accounts/apps.py` so that `AnonymousUser.is_admin` always returns `False`, preventing AttributeError when code checks `request.user.is_admin` for unauthenticated users.

**Testing improvements:**
* Added a test in `apps/accounts/tests/test_models.py` to confirm that `AnonymousUser.is_admin` returns `False` and does not raise an error.
* Added tests for provider and label autocomplete endpoints to ensure unauthenticated users receive empty results and no 500 errors are raised.
* Added a test to ensure that an anonymous user accessing the explorer index page is shown the login page without a 500 error.
* Added a test to verify that anonymous users attempting to access the provider edit URL are redirected to login and do not cause a 500 error.

**Test setup:**
* Imported `AnonymousUser` in `apps/accounts/tests/test_models.py` for use in new tests.

### DIFF
--- a/apps/accounts/apps.py
+++ b/apps/accounts/apps.py
@@ -3,3 +3,10 @@ from django.apps import AppConfig
 
 class AccountsConfig(AppConfig):
     name = "apps.accounts"
+
+    def ready(self):
+        # Patch AnonymousUser so that code accessing `request.user.is_admin`
+        # does not raise AttributeError for unauthenticated users.
+        from django.contrib.auth.models import AnonymousUser
+
+        AnonymousUser.is_admin = property(lambda self: False)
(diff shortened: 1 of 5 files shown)


mrchrisadams:
@timcowlishaw When I was chatting to Fershad this morning, I discovered a couple of 500's being triggered. This PR fixes them. Tagging you so you know it's gone through and what it was for.

~~Ship~~ / **Show** / ~~Ask~~

